Issue #3, opened 2019-06-27
Implement get_storage_client


exarkun:
Implement a client for #2 that supplies tokens to remote calls.

Don't implement the crypto. That comes later.

Issue #2, opened 2019-06-16
Implement get_storage_server


exarkun:
Offer a storage server that requires tokens as a parameter.

Don't implement the crypto.  That comes later.

PR #4, opened 2019-07-05
Cryptoless Tahoe-LAFS storage server plugin

exarkun:
Fixes: #2 
Fixes: #3 

This seems to work (that is, it allows the same functionality as native Tahoe-LAFS):

```
$ tahoe -d client/ create-alias tahoe
Alias 'tahoe' created
$ tahoe -d client/ list-aliases
  tahoe: URI:DIR2:itgd...
$ tahoe -d client/ ls tahoe:
$ tahoe -d client/ put ~/.profile tahoe:.profile
201 Created
URI:CHK:mi3f...:1:1:477
$ diff <(tahoe -d client/ get tahoe:.profile) ~/.profile
$ 
```

It might make sense to pull the Nix parts of this out to a separate PR.  I'm happy to do that if it helps the review process.


crwood:
I should have maybe mentioned above that I did not test your nix expressions. :)

src/_secureaccesstokenauthorizer/_plugin.py
```diff
@@ -0,0 +1,78 @@
...
+
+@implementer(IFoolscapStoragePlugin)
+class SecureAccessTokenAuthorizer(object):
+    """
+    A storage plugin which provides a token-based access control mechanism on
+    top of the Tahoe-LAFS built-in storage server interface.
+    """
+    name = u"privatestorageio-satauthz-v1"
+
+    def get_storage_server(self, configuration, get_anonymous_storage_server):
```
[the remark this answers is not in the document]
exarkun: > Consider omitting `self` (since it is unused here) and using the `@staticmethod` decorator.

Hm, possible, yea, but I'm not sure much is gained.  In this case, I think `self` may eventually be used as the server will need to know more in order to validate and then burn tokens.

src/_secureaccesstokenauthorizer/_plugin.py
```diff
@@ -0,0 +1,78 @@
...
+        )
+        return succeed(
+            AnnounceableStorageServer(
+                announcement,
+                storage_server,
+            ),
+        )
+
+
+    def get_storage_client(self, configuration, announcement, get_rref):
```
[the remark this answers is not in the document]
exarkun: Added a minimal test that at least makes sure it returns the right kind of object.  `configuration` and `announcement` will be used later.

src/_secureaccesstokenauthorizer/tests/strategies.py
```diff
@@ -0,0 +1,208 @@
...
+
+def shares():
+    """
+    Build Tahoe-LAFS share data.
+    """
+    return tuples(
+        sharenums(),
+        sizes()
+    ).map(
+        lambda (num, size): bytes_for_share(num, size),
```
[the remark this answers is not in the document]
exarkun: They are necessary but they are also a Python 3 syntax error, indeed.  Fixing.

src/_secureaccesstokenauthorizer/tests/test_plugin.py
```diff
@@ -0,0 +1,178 @@
...
+        """
+        The plugin can be discovered.
+        """
+        self.assertThat(
+            getPlugins(IFoolscapStoragePlugin),
+            Contains(storage_server),
+        )
+
+
+    def test_provides_interface(self):
```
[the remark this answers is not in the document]
exarkun: Hm!  Maybe.  I don't know what `TestCase` things of having staticmethods.  I think it's idiomatic to have test methods as instance methods and it's more maintainable and composable to have little or no use of `self` in such methods.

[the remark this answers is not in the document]
exarkun: I should switch this to use `assertThat(..., Provides(...))` though.

tahoe-lafs.nix
```diff
@@ -0,0 +1,32 @@
+{ nettools, pythonPackages, buildPythonPackage, eliot }:
+buildPythonPackage rec {
+  version = "1.14.0.dev";
+  name = "tahoe-lafs-${version}";
+  src = ~/Work/python/tahoe-lafs;
```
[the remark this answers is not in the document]
exarkun: Nope!

src/_secureaccesstokenauthorizer/_storage_client.py
```diff
@@ -0,0 +1,148 @@
...
+    IStorageServer,
+)
+
+@implementer(IStorageServer)
+@attr.s
+class SecureAccessTokenAuthorizerStorageClient(object):
+    """
+    XXX
+    """
+    _get_rref = attr.ib()
```
meejah: Docstring, please.
(Also as a general thing, this is a "private" module with an underscore inside another private module w/ underscore, with underscored `attr.ib()`s -- does it need all that "private-ness" or is this a convention in here..?)

src/_secureaccesstokenauthorizer/_storage_server.py
```diff
@@ -0,0 +1,169 @@
...
+
+    :return foolscap.remoteinterface.RemoteMethodSchema: The new schema
+        object.
+    """
+    new_kwargs = dict(schema.argConstraints)
+    new_kwargs.update(kwargs)
+    modified_schema = RemoteMethodSchema(**new_kwargs)
+    # Initialized from **new_kwargs, RemoteMethodSchema.argumentNames is in
+    # some arbitrary, probably-incorrect order.  Fix it.
+    modified_schema.argumentNames = (
```
meejah: Instead of "fix it", perhaps describe the desired order?
...and I guess this can all go away in the glorious 3.7+ future..? :/

src/_secureaccesstokenauthorizer/_storage_server.py
```diff
@@ -0,0 +1,169 @@
...
+        list(argName for (argName, _) in kwargs) +
+        schema.argumentNames
+    )
+    return modified_schema
+
+
+
+class RITokenAuthorizedStorageServer(RemoteInterface):
+    __remote_name__ = (
```
meejah: docstring please

src/_secureaccesstokenauthorizer/_storage_server.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+
+@implementer_only(RITokenAuthorizedStorageServer, IReferenceable, IRemotelyCallable)
+# It would be great to use `frozen=True` (value-based hashing) instead of
+# `cmp=False` (identity based hashing) but Referenceable wants to set some
+# attributes on self and it's hard to avoid that.
+@attr.s(cmp=False)
+class SecureAccessTokenAuthorizerStorageServer(Referenceable):
+    _original = attr.ib()
```
meejah: docstring please (This is basically a facade that calls through to the "real"/original storage-server API, but validates tokens first?)

src/_secureaccesstokenauthorizer/_storage_server.py
```diff
@@ -0,0 +1,169 @@
...
+    def remote_slot_testv_and_readv_and_writev(self, tokens, *a, **kw):
+        self._validate_tokens(tokens)
+        return self._original.remote_slot_testv_and_readv_and_writev(*a, **kw)
+
+    def remote_slot_readv(self, *a, **kw):
+        return self._original.remote_slot_readv(*a, **kw)
+
+# I don't understand why this is required.
+# SecureAccessTokenAuthorizerStorageServer is-a Referenceable.  It seems like
+# the built in adapter should take care of this case.
```
meejah: Is it because you said `implementer_only` instead of `implementer`...? (also just guessing)

src/_secureaccesstokenauthorizer/tests/strategies.py
```diff
@@ -0,0 +1,208 @@
...
+    return dictionaries(
+        sharenums(),
+        test_and_write_vectors(),
+        # An empty dictionary wouldn't make much sense.  And it provokes a
+        # NameError from Tahoe, storage/server.py:479, `new_length` referenced
+        # before assignment.
+        min_size=1,
+        # Just for practical purposes...
+        max_size=8,
+    )
```
meejah: I wonder if these should have names that don't start with `test_` just in case there's either human or machine confusion about whether these are (supposed to be) tests...

src/_secureaccesstokenauthorizer/_storage_server.py
exarkun: yep

src/_secureaccesstokenauthorizer/_storage_server.py
exarkun: oh yea maybe because I didn't include `providedBy(Referenceable)` in the list of things it only implements...

exarkun: `providedBy(Referenceable())` or `implementedBy(Referenceable)`, of course.  Or maybe `directlyProvidedBy(Referenceable())` / `directlyImplementedBy(Referenceable)`.

exarkun: Apparently there is no `directlyImplementedBy` and adding `directlyProvidedBy(Referenceable())` to the `implements_only` list doesn't make a difference.  So I dunno.

src/_secureaccesstokenauthorizer/tests/strategies.py
exarkun: Hm.  Reasonable.  I wonder what the name should be, though.

Issue #5, opened 2019-07-08
Set up CI for testing


exarkun:
There's an automated test suite.  Run it automatically!

Since we're living on the edge and importing from Tahoe-LAFS python packages we should test against Tahoe-LAFS master@HEAD somehow.  That might be _all_ we do for now since this code is definitely broken and useless against all released versions of Tahoe-LAFS.

exarkun:
CircleCI is set up to do some things.  It hasn't been posting status updates to GitHub because I revoked its github API access.  It doesn't need this access to do builds.  It does need this access to post build results ("write commit status").  GitHub describes the exact access that CircleCI has requested in this impressively vague way:

> ... ability to request access to private data ...

I can't find any GitHub UI that explains exactly what access CircleCI has requested.  CircleCI documentation claims:

> Read Permission
> * Get a user’s email address
>
> Write Permissions
> * Add deploy keys to a repo
> * Add service hooks to a repo
> * Get a list of a user’s repos
> * Add an SSH key to a user’s account

Leaving aside the fact that "Get a list of a user's repos" is clearly a read permission, this specification is also vague because, I suppose, CircleCI is trying to keep it generic across Bitbucket and GitHub.  Also, we only have CircleCI's word that this is what they actually request (it could be simply out of date with respect to their actual service, not necessarily maliciously incorrect).

I guess I am going to let it have API access again because there don't seem to be any better alternatives and everything on the PrivateStorageio org is public right now anyway.  This is certainly open to reconsideration at any future point.

exarkun:
Fixed partly by clicking PrivateStorageio org GitHub buttons and partly by cleaning up the existing circleci config as part of #4

PR #6, opened 2019-07-09
Tahoe-LAFS is not Python 3 yet so neither can we be.

exarkun:
Fixes: #5


exarkun:
This is maybe silly and just needs to be part of #2.  Closing this for now on that basis.

PR #118, opened 2020-04-03
Bust the cache key

### LINKED ISSUE #117: Linux CircleCI tests job takes a lot longer than necessary
This job rebuilds a lot of dependencies every time it runs.  These could be cached, instead.

exarkun:
Our dependencies changed a while ago but we probably kept using the same cache
so we never updated it with the builds of the new dependencies.

Fixes #117


exarkun:
This speeds it up slightly, at least.  python-challenge-bypass-ristretto is now in the cache.  The test suite itself currently takes more than a little time so this isn't as a big of a win as I hoped.  Still, a bit of a win.

Maybe in the future we can compute cache keys better so that we don't need to do this kind of manual bumping.

Issue #97, opened 2020-03-06
The codecov reports are missing


exarkun:
CI is configured to collect coverage information from the test suite and submit it to codecov.  However, a look at the codecov badge or the page for the project reveals no coverage information.

PR #139, opened 2020-05-01
Compute >= 2 GiB share size correctly

### LINKED ISSUE #138: ZKAP requirement is miscalculated for shares larger than 2 GiB
Note that "shares" are storage for immutable files (storage for mutable files is in "slots").  This issue applies to shares.

Number of ZKAPs required for some amount of storage is based on that amount.  The server-side code that determines the current size of a share is probably incorrect for files larger than 2 GiB.  It reads size data from bytes 4-7 of the share file.  However, the maximum size these bytes can represent is 2 GiB.  When the share is larger, the real size has to be computed from the size of the container (the file) itself by subtracting some bookkeeping overhead from the total size.

exarkun:
Fixes #138


exarkun:
The missing coverage seems to be from not collecting/integrating coverage information from Linux runs on CI.  Only the Windows coverage is being reported.  This is probably related to #97 .

Issue #142, opened 2020-05-06
Disable Eliot test suite


exarkun:
It has intermittent spurious failures.

PR #145, opened 2020-05-06
Pin Tahoe-LAFS to 1.14.0

### LINKED ISSUE #143: Pin Tahoe-LAFS 1.14.0 on CI
We use the Tahoe-LAFS Python API but there is no guaranteed stable/public API.  To avoid spurious breakage due to upstream Tahoe-LAFS changes, pin a known version and only upgrade it when we're sure we want to.

exarkun:
Fixes #143


exarkun:
I ended up doing this pinning in #144 as I was fixing #142

PR #178, opened 2020-09-14
Increase PaymentController RETRY_INTERVAL to 1 sec

### LINKED ISSUE #177: The aggressive RETRY_INTERVAL in PaymentController consumes enormous CPU resources
https://github.com/PrivateStorageio/ZKAPAuthorizer/commit/cb534cc1d465a5ba8b414faa5a9d41506e59beda reduced the retry timer used by controller.PaymentController._schedule_retries to 0 seconds (while a later commit, https://github.com/PrivateStorageio/ZKAPAuthorizer/commit/6343ce3885467e63bf1baf52d88e48c1e5eff85f, increased this value to 1 milliseconds) . On My Machine, this very short value causes otherwise idle `tahoe` processes to sustain 100% CPU usage indefinitely.

Increasing `RETRY_INTERVAL` to 1000 milliseconds brings CPU-usage levels back down to less-knee-scorching single-digit percentages, though I'm unsure whether this constitutes a reasonable adjusted value (and/or whether changing it would have any adverse effects on other parts of the system -- beyond introducing an extra-second wait-time at the beginning of redemption-attempts). I'm happy to submit a PR for this, in any case; please advise what an appropriate value might be.

crwood:
Fixes #177

### DIFF
--- a/src/_zkapauthorizer/controller.py
+++ b/src/_zkapauthorizer/controller.py
@@ -102,7 +102,7 @@ from .model import (
     Error as model_Error,
 )
 
-RETRY_INTERVAL = timedelta(milliseconds=1)
+RETRY_INTERVAL = timedelta(milliseconds=1000)
 
 @attr.s
 class UnexpectedResponse(Exception):


exarkun:
Okay, I finally understand what's going on here.  Sorry it took me so long.  I still want to apply a more complete fix here but this can certainly be merged to mitigate the issue until then and to let any downstream projects that might be carrying a patch to work-around this issue stop doing so. :)

hacklschorsch:
👍

PR #188, opened 2021-05-06
Eliminate white space around ristretto key when reading key file

hacklschorsch:
It is easy to erroneously add extra white space to a text file. For example when writing with an editor like `nano` or when using `echo` without the `-n` argument.

Handing that white space over to python-challenge-bypass-ristretto's `decode_base64()` method will make it fail in a rather opaque way, see https://github.com/LeastAuthority/python-challenge-bypass-ristretto/issues/37.

This eliminates white space including newlines around the key expected to be in that file.

### DIFF
--- a/src/_zkapauthorizer/_plugin.py
+++ b/src/_zkapauthorizer/_plugin.py
@@ -145,7 +145,7 @@ class ZKAPAuthorizer(object):
         signing_key = SigningKey.decode_base64(
             FilePath(
                 kwargs.pop(u"ristretto-signing-key-path"),
-            ).getContent(),
+            ).getContent().strip(),
         )
         announcement = {
             u"ristretto-issuer-root-url": root_url,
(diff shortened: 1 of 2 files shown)


exarkun (commented):
Thanks.  Do you feel up to writing a unit test, too?

hacklschorsch:
I tried for one hour to run the existing tests, then gave up.

exarkun:
Yea, fair :(  I also started looking at the state of the test suite and it's not great.  I'll loop back here after it has been revivified...

hacklschorsch:
I could also make the point that Unit testing here might not be the right thing:

  - the `SigningKey.decode_base64()` unit is imported from another library and tested there. It specifies the parameter it expects, and it's our job to adhere to the spec
  -  the `FilePath(...).getContent()` unit is from the standard library and widely tested,
  - as is the `strip()` unit.

It's the integration of all this we want to test.
Testing that integration makes good sense, and I also think it is a good location to spec our expected input.

I think it is worth while to add an integration test for this.

But honestly, trying to be very street smart I want to do something else.

hacklschorsch:
That violates the principle that Unit tests should test one thing only. But OTOH, if some other component has a problem with a signing key like that, it should be exposed as well.

exarkun:
In #198 I took a shot at the kind of "unit test" I think provides value here.  I started w/ this PR for that branch so I'm gonna say that PR supersedes this one.

PR #192, opened 2021-05-12
Try the other URL

### LINKED ISSUE #191: macOS CI broken because get-pip.py URL changed
```
#!/bin/bash --login -eo pipefail
# The CircleCI macOS environment has curl and Python but does not
# have pip.  So, for starters, use curl and Python to get pip.
if [ "2.7" == "2.7" ]; then
  curl https://bootstrap.pypa.io/2.7/get-pip.py -o get-pip.py
else
  curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
fi
python2.7 get-pip.py

  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   936  100   936    0     0   7800      0 --:--:-- --:--:-- --:--:--  7800

Hi there!

The URL you are using to fetch this script has changed, and this one will no
longer work. Please use get-pip.py from the following URL instead:

    https://bootstrap.pypa.io/pip/2.7/get-pip.py

Sorry if this change causes any inconvenience for you!

We don't have a good mechanism to make more gradual changes here, and this
renaming is a part of an effort to make it easier to us to update these
scripts, when there's a pip release. It's also essential for improving how we
handle the `get-pip.py` scripts, when pip drops support for a Python minor
version.

There are no more renames/URL changes planned, and we don't expect that a need
would arise to do this again in the near future.

Thanks for understanding!

- Pradyun, on behalf of the volunteers who maintain pip.
```

exarkun:
Fixes #191

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -74,7 +74,8 @@ jobs:
             # The CircleCI macOS environment has curl and Python but does not
             # have pip.  So, for starters, use curl and Python to get pip.
             if [ "<< parameters.py-version >>" == "2.7" ]; then
-              curl https://bootstrap.pypa.io/2.7/get-pip.py -o get-pip.py
+
+              curl https://bootstrap.pypa.io/pip/2.7/get-pip.py -o get-pip.py
             else
               curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
             fi


exarkun (commented):
The macOS jobs are green now.

PR #194, opened 2021-05-12
Try removing the vcpython27 installation step

### LINKED ISSUE #193: Windows CI broken because Microsoft stopped hosting vcpython27
It's not possible to install vcpython27 from chocolatey anymore because Microsoft stopped hosting one of the artifacts.  Thus CI fails on Windows at the step where we try to install vcpython27.

Maybe we don't need it?  There's no extension code in ZKAPAuthorizer... But if any dependency needs to be built then we're somewhat out of luck (I imagine netifaces will hit us like it hit Tahoe-LAFS since Tahoe-LAFS is a dependency...)

exarkun:
Fixes #193

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -24,12 +24,6 @@ jobs:
       run: |
         git fetch --tags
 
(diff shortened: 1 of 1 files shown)


exarkun:
The Windows job is green now.

PR #196, opened 2021-05-12
Remove the bundled Tahoe-LAFS and just require it as a param or nixpkgs attr

### LINKED ISSUE #195: Delegate Nix packaging of other software to someone else
Presently the Nix packaging for ZKAPAuthorizer also includes hooks for Tahoe-LAFS nix packaging.  The Tahoe-LAFS packaging is somewhat complicated by the need to work around a couple shortcomings in the upstream Tahoe-LAFS packaging.

The complications then create further complications for anyone packaging ZKAPAuthorizer.  If ZKAPAuthorizer consumers are going to have their lives complicated, we may as well not have the complications in ZKAPAuthorizer itself.

Instead of what we do now, just assume there is a Tahoe-LAFS in nixpkgs that we can use or that the caller will supply one directly.

This specifically makes it easier to integrate with PrivateStorageio/nixpkgs which does have a good version of Tahoe-LAFS and which the rest of the PrivateStorageio stack wants to use anyway.

exarkun:
Fixes #195


exarkun:
NixOS build now passing.

PR #198, opened 2021-05-17
Signing key whitespace handling

### LINKED ISSUE #197: Extraneous whitespace around the signing key results in an opaque failure
This case leads to a `DecodeException` from the underlying library.  It's roughly like:

```
  File "challenge_bypass_ristretto/__init__.py", line 61, in decode_base64
    raise DecodeException()
challenge_bypass_ristretto.DecodeException
```

This is the same exception as any key decoding problem provokes.  Since the key itself is quite opaque (not quite a purely random byte string but close) it's far from obvious that this signals a whitespace problem.  Debugging the failure then involves serious digging or making a lucky guess or bumping into someone who has already figured this out.

The underlying library could be more informative in its exceptions but, failing that, ZKAPAuthorizer can likely improve the user experience for this scenario.

exarkun:
Fixes #197 

Based on #188 

This adds a new helper, `load_signing_key`, and makes it responsible for whitespace cleanup.  Then it adds a test that exercises several cases of `load_signing_key` directly so the failures are more easily directly attributable to the offending code.


hacklschorsch:
Sweet!

PR #202, opened 2021-06-03
Give the Ristretto build step a longer no-output timeout

### LINKED ISSUE #200: The CircleCI cache frequently expires and forces a massive rebuild of our dependencies
Since we have no Nix binary cache set up we have to build a huge amount of software just to get to the point where we can run the ZKAPAuthorizer test suite.  We have a CircleCI cache set up for these dependencies.  Unfortunately CircleCI deletes it after 15 days.  Development is not active enough to make sure this cache gets refreshed within that window (which avoids the deletion and the need to rebuild everything).

Run a job on a schedule that refreshes the cache so we only have to rebuild our deps if they actually need rebuild, not just because the cache got dropped.

exarkun:
Rust stuff takes a while to build I guess.

Fixes #200

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -200,6 +200,7 @@ jobs:
 
       - run:
           name: "Build challenge-bypass-ristretto"
+          no_output_timeout: "90m"
           command: |
             # Pre-build this because doing so is somewhat memory intensive and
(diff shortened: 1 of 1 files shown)


exarkun:
This has been fixed by using Cachix instead of the CircleCI cache, instead.

PR #204, opened 2021-06-11
Bump tahoe dependency to 1.15.1.

tomprince:


exarkun:
I think this will be obsoleted by #229

PR #208, opened 2021-06-14
Set up Cachix cache

### LINKED ISSUE #200: The CircleCI cache frequently expires and forces a massive rebuild of our dependencies

exarkun:
Fixes #200

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -135,11 +135,15 @@ jobs:
             ZKAPAUTHORIZER_HYPOTHESIS_PROFILE: "ci"
 
   linux-tests: &LINUX_TESTS
+    resource_class: "xlarge"
     docker:
       # Run in a highly Nix-capable environment.
       - image: "nixorg/nix:circleci"
 
     environment:
+      # CACHIX_AUTH_TOKEN is manually set in the CircleCI web UI and allows us to push to CACHIX_NAME.
+      CACHIX_NAME: "privatestorage-opensource"
+
       # Specify a revision of PrivateStorageio/nixpkgs to run against.  This
       # essentially pins the majority of the software involved in the build.
       # This revision is selected arbitrarily (it's just new enough to define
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -234,28 +180,18 @@ jobs:
               --arg collectCoverage true \
               --attr doc
```
tomprince, line 182: (via https://circleci.com/blog/managing-secrets-when-you-have-pull-requests-from-outside-contributors/)
```suggestion

      - run:
          name: Early return if this build is from a forked PR
          command: |
            if [ -n "$CIRCLE_PR_NUMBER" ]; then
              echo "Nothing to do for forked PRs, so marking this step successful"
              circleci step halt
            fi
            
```

I guess this and the push need to move after the coverage reporting, though.

tomprince (approved):
This looks good. It would be good to prevent trying to push to cachix from PRs, though.

Did you populate the cache from a local machine? (I'm curious how you avoided the timeout building everything for the cache)

exarkun:
> This looks good. It would be good to prevent trying to push to cachix from PRs, though.

Hm.  Thinking out loud for a moment...

We might roughly divide what's being pushed into two categories.  Our dependencies and our stuff (with only ZKAPAuthorizer in the latter category, really).  It is useful to push our dependencies in all cases because it means any subsequent build can re-use them without building them (even if we had to build them ourselves the first time).  It might still be useful to have PRs push our dependencies so that only the first build of any PR which changes our dependencies has to build them.

If we consider our stuff instead, the point is not to avoid a rebuild on CI.  However, developer/deployment systems could be set up to use Cachix as well and then those systems can avoid building our stuff.  They might want to use a build from a PR, I suppose, for manual testing purposes?  More useful might be builds from master, or even release tags.

> 
> Did you populate the cache from a local machine? (I'm curious how you avoided the timeout building everything for the cache)

It was populated from a CI run.  I don't really understand this myself.  Somehow with the configuration change builds began succeeding in one hour (sans cache) instead of timing out after two or three.

I have no idea if this is because something changed on CircleCI's side or if it's something we did.  As a cheap (but probably not fast) experiment I've restarted a recent master build that timed out - https://app.circleci.com/pipelines/github/PrivateStorageio/ZKAPAuthorizer/631/workflows/027b0bb7-e9a7-43fb-8e7e-0d077d45ee2b/jobs/1459.  If it times out again then that points towards a change in the branch improving things.  If it passes then I think that points towards CircleCI changing something on their end.

The only change in the branch that really seems like it could have made a difference (but also seems like it *should not* have made this difference) is the change in Docker image being used, from `nixorg/nix:circleci` to `nixos/nix:latest`.  Maybe Nix in one of those images had some horrible performance problem?  Either way all of the stuff *we* build is the same.  If a Nix difference doesn't explain it, maybe this some crazy performance-is-randomly-5x-worse-because-this-one-Docker-image-quirk cases...


I don't really understand

exarkun:
> > This looks good. It would be good to prevent trying to push to cachix from PRs, though.
> 
> Hm. Thinking out loud for a moment...
> 

Er, unless you were just referencing the inline comment regarding *forked* PRs here... then nevermind.

tomprince:
> Er, unless you were just referencing the inline comment regarding forked PRs here... then nevermind.

I've found CircleCI's documentation to be somewhat opaque but [this](https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables) *does* suggests that is only on forked PRs. (I had also read [this](https://support.circleci.com/hc/en-us/articles/360006357533--CIRCLE-PULL-REQUEST-is-missing-from-my-Pull-Request) which made me wonder initially if it would also be set for in-repo PRs, but that is a different environment variable

.circleci/config.yml
```diff
@@ -234,28 +180,30 @@ jobs:
...
+              # I'm sure you're thinking "CIRCLE_PR_NUMBER must just be the
+              # number of the PR being built".  Sorry, dear reader, you have
+              # guessed poorly.  It is also conditionally set based on whether
+              # this is a PR from a fork or not.
+              #
+              # https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables
+              echo "Skipping Cachix push for forked PR."
+            else
+              # https://docs.cachix.org/continuous-integration-setup/circleci.html
+              bash -c "comm -13 <(sort /tmp/store-path-pre-build | grep -v '\.drv$') <(nix path-info --all | grep -v '\.drv$' | sort) | cachix push $CACHIX_NAME"
```
hacklschorsch, line 199: Sexy.

This uploads all the things, right? I am looking forward to see how it works out with the LRU caching and the 5 GB :)

.circleci/config.yml
```diff
@@ -137,85 +137,31 @@ jobs:
   linux-tests: &LINUX_TESTS
     docker:
       # Run in a highly Nix-capable environment.
-      - image: "nixorg/nix:circleci"
+      - image: "nixos/nix:latest"
```
hacklschorsch, line 140: Nice!

hacklschorsch (commented):
How can I say that I like this and think it can be merged, i.e. a green little check mark please

PR #211, opened 2021-07-21
Make DummyRedeemer require a public key

### LINKED ISSUE #210: Require a public key in DummyRedeemer
DummyRedeemer allows `None` as a value for its public key.  It is never legitimate for a real issuer to not offer us a public key.  Doing so completely breaks the security model for the client (they lose the ability to prove the issuer is not tracking them by using a custom signing key).

So, make DummyRedeemer more like real usage and have it always have a public key.

This is motivated by #110 where the client is going to start looking at the public key more carefully and at best it would be a lot of unnecessary complexity to handle `None` public key from `DummyRedeemer`.

exarkun:
Fixes #210

### DIFF
--- a/src/_zkapauthorizer/tests/test_client_resource.py
+++ b/src/_zkapauthorizer/tests/test_client_resource.py
@@ -192,6 +192,32 @@ from .json import (
 
 TRANSIENT_ERROR = u"something went wrong, who knows what"
 
+def get_dummyredeemer_public_key(plugin_name, node_config):
+    """
+    Get the issuer public key a ``DummyRedeemer`` has been configured with.
+
+    :param unicode plugin_name: The plugin name to use to choose a
+        configuration section.
+
+    :param _Config node_config: See ``from_configuration``.
+    """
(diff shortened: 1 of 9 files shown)


src/_zkapauthorizer/tests/test_client_resource.py
```diff
@@ -192,6 +192,32 @@
...
+    :param _Config node_config: See ``from_configuration``.
+    """
+    section_name = u"storageclient.plugins.{}".format(plugin_name)
+    redeemer_kind = node_config.get_config(
+        section=section_name,
+        option=u"redeemer",
+    )
+    if redeemer_kind != "dummy":
+        raise ValueError(
+            "Cannot read dummy redeemer public key from configuration for {!r} redeemer.".format(
```
exarkun, line 211: I'd write a test for this but turns out there are no callers of this function in the follow-up PR that actually does something useful so I'll just delete the whole thing, instead.

hacklschorsch (commented):
Nice, nice, very nice

PR #212, opened 2021-07-22
Close the trust loop

### LINKED ISSUE #110: Defend against a key shenanigans attack by the server
The privacy features of challenge-bypass-ristretto are not guaranteed if the client cannot hide in a crowd of other users.  The size of the crowd is defined by the number of clients which redeem vouchers from a server using a particular signing key.  If a server offers a different signing key to each user, no user has a crowd and there are no privacy features.

Outside of the cryptographic protocol, the client must take steps to be sure it is not being attacked this way.  These steps must somehow convince the client that the server is signing a large number of tokens with the same key.

- [ ] [Identify suspicious ZKAPs](https://github.com/PrivateStorageio/ZKAPAuthorizer/issues/112)
- [ ] [Don't spend suspicious ZKAPs](https://github.com/PrivateStorageio/ZKAPAuthorizer/issues/113)
- [ ] [Expose suspicious ZKAPs in the HTTP API](https://github.com/PrivateStorageio/ZKAPAuthorizer/issues/114)

exarkun:
Fixes #110

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -52,6 +52,20 @@ The client can also be configured with the number of passes to expect in exchang
 
 The value given here must agree with the value the issuer uses in its configuration or redemption may fail.
 
+allowed-public-keys
+~~~~~~~~~~~~~~~~~~~
+
+Regardless of which redeemer is selected,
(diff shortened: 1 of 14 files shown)


hacklschorsch (approved):
All clear!

PR #214, opened 2021-07-29
Proof-of-concept implementation of lease renewal and price calculation using tahoe's iteration.

tomprince:
This is completely untested. It isn't obvious if the code can use the existing fakes, and I don't have setup to test live.

Also, the code is *really* messy, though it'd be straightforward to refactor.

(Incidentally, I think I may know how to hook into the operations stuff, if we want to present the price result that way, rather than waiting)

### DIFF
--- a/src/_zkapauthorizer/lease_maintenance.py
+++ b/src/_zkapauthorizer/lease_maintenance.py
@@ -49,7 +49,6 @@ from twisted.python.log import (
 )
 
 from allmydata.interfaces import (
-    IDirectoryNode,
     IFilesystemNode,
 )
 from allmydata.util.hashutil import (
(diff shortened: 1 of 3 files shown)


meejah:
The main trade-off to consider here I think is that doing a traversal like this will be 100% accurate -- but it will cost a whole bunch of round-trips to the storage servers (linear with the number of Snapshots) versus e.g. "estimating" based on the fact that the Snapshot cap and the metadata cap will each be less than 1MB.

I put "estimate" in scare-quotes because we _do_ know that a Snapshot and the metadata are both definitely less than 1MB (and if we ever come anywhere near that size, we could do something else at that point). For example, we could put in unit-tests etc in magic-folder that fail if either one is more than 500KB.

exarkun:
Possibly a good idea but not something we plan to pursue in the near future.  Just closing so that the list of open PRs reflects active development efforts

PR #216, opened 2021-08-02
Switch to fully using the 31 day lease time

### LINKED ISSUE #215: ZKAPs are spent faster than the value of 1MB×month might imply
The lease maintenance service will renew any lease it encounters which is due to expire in 3 days or less.  Since lease renewal involves setting the lease expiration time to the current time plus 31 days, this means that the practical duration of a lease may be as small as (31 days - 3 days = ) 28 days.

This is confusing.  It is reasonable to summarize 31 days as "one month".  It is probably even reasonable to summarize 30.5 days as "one month".  28 days is not a reasonable definition of a month.

Ultimately the renewal scheme should be replaced by a system that does not throw away part of the lease period.  In the interim, we can change the lease maintenance service to let leases fully expire before renewing them.  On a system with garbage collection enabled this would be somewhat dangerous but ZKAPAuthorizer is not yet in shape to do that anyway.

exarkun:
Fixes #215

### DIFF
--- a/src/_zkapauthorizer/_plugin.py
+++ b/src/_zkapauthorizer/_plugin.py
@@ -266,7 +266,19 @@ def _create_maintenance_service(reactor, node_config, client_node):
         get_root_nodes=partial(get_root_nodes, client_node, node_config),
         storage_broker=client_node.get_storage_broker(),
         secret_holder=client_node._secret_holder,
-        # Make this configuration
+        # The greater the min lease remaining time, the more of each lease
+        # period is "wasted" by renewing the lease before it has expired.  The
+        # premise of ZKAPAuthorizer's use of leases is that if they expire,
+        # the storage server is free to reclaim the storage by forgetting
+        # about the share.  However, since we do not know of any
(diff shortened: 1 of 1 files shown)


src/_zkapauthorizer/_plugin.py
```diff
@@ -266,7 +266,19 @@ def get_now():
...
+        # the storage server is free to reclaim the storage by forgetting
+        # about the share.  However, since we do not know of any
+        # ZKAPAuthorizer-enabled storage grids which will garbage collect
+        # shares when leases expire, we have no reason not to use a zero
+        # duration here - for now.
+        #
+        # In the long run, storage servers must run with garbage collection
+        # enabled.  Ideally, before that happens, we will have a system that
+        # doesn't involve trading of wasted lease time against reliability of
+        # leases being renewed before the shares are garbage collected.
```
hacklschorsch, line 281: 🤞

hacklschorsch (approved):
🎵

Issue #224, opened 2021-09-07
Broken against recent Tahoe-LAFS master@HEAD


exarkun:
```
Unhandled Error
Traceback (most recent call last):
  File "/nix/store/2jsiq3gnn0b61kzvmjqmwpqm0al8k95g-python2.7-Twisted-18.9.0/lib/python2.7/site-packages/twisted/scripts/trial.py", line 478, in postOptions                                 
    _BasicOptions.postOptions(self)
  File "/nix/store/2jsiq3gnn0b61kzvmjqmwpqm0al8k95g-python2.7-Twisted-18.9.0/lib/python2.7/site-packages/twisted/scripts/trial.py", line 388, in postOptions                                 
    self['reporter'] = self._loadReporterByName(self['reporter'])
  File "/nix/store/2jsiq3gnn0b61kzvmjqmwpqm0al8k95g-python2.7-Twisted-18.9.0/lib/python2.7/site-packages/twisted/scripts/trial.py", line 375, in _loadReporterByName                         
    for p in plugin.getPlugins(itrial.IReporter):
  File "/nix/store/2jsiq3gnn0b61kzvmjqmwpqm0al8k95g-python2.7-Twisted-18.9.0/lib/python2.7/site-packages/twisted/plugin.py", line 213, in getPlugins                                         
    allDropins = getCache(package)
--- <exception caught here> ---
  File "/nix/store/2jsiq3gnn0b61kzvmjqmwpqm0al8k95g-python2.7-Twisted-18.9.0/lib/python2.7/site-packages/twisted/plugin.py", line 171, in getCache                                           
    provider = pluginModule.load()
  File "/nix/store/2jsiq3gnn0b61kzvmjqmwpqm0al8k95g-python2.7-Twisted-18.9.0/lib/python2.7/site-packages/twisted/python/modules.py", line 392, in load                                       
    return self.pathEntry.pythonPath.moduleLoader(self.name)
  File "/nix/store/2jsiq3gnn0b61kzvmjqmwpqm0al8k95g-python2.7-Twisted-18.9.0/lib/python2.7/site-packages/twisted/python/reflect.py", line 308, in namedAny                                   
    topLevelPackage = _importAndCheckStack(trialname)
[...]

PR #226, opened 2021-09-15
Use `add_lease` instead of `renew_lease`

### LINKED ISSUE #220: Use `add_lease` instead of `renew_lease` everywhere.
As far as I can tell, `add_lease` will add or renew a lease as appropriate, but `renew_lease` will fail if a lease with our renewal secret does not exist.

I think we should:
- switch lease_maintence to call add_lease
- make `remote_renew_lease` delegate to `remote_add_lease` in our server plugin

This was also suggested upstream in https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3773

exarkun:
Fixes #220 

Maybe also #224

### DIFF
--- a/src/_zkapauthorizer/lease_maintenance.py
+++ b/src/_zkapauthorizer/lease_maintenance.py
@@ -55,6 +55,8 @@ from allmydata.interfaces import (
 from allmydata.util.hashutil import (
     file_renewal_secret_hash,
     bucket_renewal_secret_hash,
+    file_cancel_secret_hash,
+    bucket_cancel_secret_hash,
 )
 
 from .controller import (
@@ -165,6 +167,7 @@ def renew_leases(
     storage_indexes = yield iter_storage_indexes(visit_assets)
 
     renewal_secret = secret_holder.get_renewal_secret()
+    cancel_secret = secret_holder.get_cancel_secret()
     servers = list(
         server.get_storage_server()
         for server
@@ -176,6 +179,7 @@ def renew_leases(
         yield renew_leases_on_server(
             min_lease_remaining,
             renewal_secret,
+            cancel_secret,
             storage_indexes,
             server,
             activity,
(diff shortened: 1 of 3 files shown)


tomprince (approved):
This looks good to me.

I noticed that you didn't do the second part of my suggestion from #220. Looking at the code again, I guess this is because we don't have enough information (i.e. the cancel secret) to do so. I wonder if it makes sense to instead remove the method from `ZKAPAuthorizerStorageServer` and have it raise `NotImplemented` in `ZKAPAuthorizerStorageClient` (I guess the first one would have compatibility implications for our unreleased software)?

exarkun:
> This looks good to me.
> 
> I noticed that you didn't do the second part of my suggestion from #220. Looking at the code again, I guess this is because we don't have enough information (i.e. the cancel secret) to do so. I wonder if it makes sense to instead remove the method from `ZKAPAuthorizerStorageServer` and have it raise `NotImplemented` in `ZKAPAuthorizerStorageClient` (I guess the first one would have compatibility implications for our unreleased software)?

Indeed.  Thank you for following up on this point.  I think maybe at the time I just forgot about that idea.  Thinking about it now, esp. in the light of the fact that (a) Itamar discovered that there aren't actually any callers of `renew_lease` in the Tahoe-LAFS code base (apart from tests) and (b) Tahoe-LAFS has now actually dropped `renew_lease` from the network protocol entirely (ie you're not even allowed to call it anymore)...

* Against a Tahoe-LAFS <1.16 storage server a custom client *could* call our `renew_lease` so if we provide it we still need to do ZKAP authorization there.  But the standard client never does so we could also just delete our `renew_lease` implementation entirely.
* Against a Tahoe-LAFS >=1.16 storage server a custom client can no longer even get a `renew_lease` call deserialized (or whatever ... Foolscap will refuse to dispatch it) on the server.  So for this case we could also just delete our `renew_lease` implementation entirely.
* No version of the Tahoe-LAFS client ever called `renew_lease` so the client-side version doesn't matter.

So ... I'll file another ticket for deleting all of our `renew_lease` code, I think, and apparently there aren't even any compatibility implications to worry about (though I guess I'll still worry about them and test them out somehow to be sure).

PR #237, opened 2021-10-12
Add an interface for discarding some ZKAPs

### LINKED ISSUE #235: The ZKAPAuthorizer internal database contains information of some value and should be amenable to backup/recovery
If a third-party backs up the ZKAPAuthorizer database and later restores it as part of an overall failure recovery process, some of the ZKAPs in the database may already have been spent.  If they are left in place, ZKAPAuthorizer will slowly churn through them and eventually get to some unspent ZKAPs and everything will proceed as normal.

Depending on various factors (not least of which is how stale the backup is), it seems possible this churn phase could take a long, long time (a day seems likely in some plausible scenarios).  There is also no way to indicate progress towards completion during this churn since the thing ZKAPAuthorizer specifically does not know is how many ZKAPs were already spent.

Beyond ZKAPs, there is other significant state in the database that changes from time to time.  There are vouchers in various states and there are records of ZKAPs that couldn't be spent for some reason (along with a record of that reason).  All of this stuff should, ideally, be recorded in a backup so it is available after recovery.

A solution is now described in [a design document in the repo](https://github.com/PrivateStorageio/ZKAPAuthorizer/blob/01f3bcb0dc7f663df6ddcbdccb547bfa92ef2c5d/docs/source/designs/backup-recovery.rst).

The implementation tasks to realize this design:

- [x] Replicate local state to remote storage
  - [x] Allow configuration of this system
    - [x] Have a Tahoe-LAFS client API that can make a new mutable directory (#316)
    - [x] Expose an external interface for configuring it (#315)
    - [x] Locally persist state related to the replication configuration (#316, #321)
    - [x] Turn on replication in the database when the replication system is set up (#359)
    - [x] Respect configuration at startup/database-connection time and enter correct mode of operation (#328)
      - [x] Capture an event-stream to the database (https://github.com/PrivateStorageio/ZKAPAuthorizer/issues/354)
  - [x] Have a long-running operation within the Tahoe-LAFS process that coordinates snapshot and event stream creation/upload
    - [x] Have a Twisted service that is running when replication is enabled (#355)
    - [x] Periodically upload snapshot data to the replica (#374)
  - [x] Determine whether a new snapshot is required
    - [x] Detect complete absence of snapshots
      - [x] Have a Tahoe-LAFS client API that can list directory contents (#322)
    - [x] Detect snapshots that are sufficiently old that they are worth replacing
  - [x] Upload a new snapshot to remote storage
    - [x] Be able to construct a new snapshot from current state
    - [x] Have a Tahoe-LAFS client API that can upload a new object and link it into a directory (#322)
    - [x] Given some snapshot data, upload it to the mutable replica directory (#337)
  - [x] Upload new "event stream" objects to remote storage
    - [x] Have SQLite3 integration that allows an "event stream" to be captured (#328)
    - [x] Have a VoucherStore schema that allows an "event stream" to be locally persisted until it is uploaded (#349)
    - [x] Detect when an "event stream" object must be uploaded
    - [x] Extract and upload an "event stream" object
- [x] Recover from a replica
  - [x] Expose an external interface that allows the recovery process to be started (#303)
  - [x] Download a "snapshot" and replay it against a fresh database 
    - [x] Have a Tahoe-LAFS client API that can download objects given a directory cap and entry name (#298)
    - [x] Have a way to know if the local database is "fresh" enough that it is safe to proceed
    - [x] Be able to to choose the best snapshot from among a set of snapshots
    - [x] Download and apply it (#308, #309)
  - [x] Download "event streams" and safely replay them against a partially-recovered database (#403)
    - [x] Have a way to choose the event streams that apply to a chosen snapshot 
    - [x] Download and apply them

exarkun:
Fixes #235 

As I was implementing this, I did realize that this still doesn't provide a *full* backup.  "Spent" is one terminal state for unblinded tokens but it isn't the only one.  "Some spending error" is also possible.  The checkpoint scheme here will not keep such tokens backed up.  Only the full copy of the database, done after each purchase, will cause that state to be backed up.

In other words, if some tokens cannot be spent for some reason, that reason will be lost unless a purchase-triggered full backup happens after the fact.

There are similar issues with other non-unblinded-token state.  The lease maintenance spending table will not be kept up to date by the checkpoint either, for example.

Despite that, this is still a small incremental improvement (it allows a third-party like GridSync to at least back up the whole database *sometimes* whereas currently it never does) so I still think it should land.  I'm filing other tickets to get us *all* the way to a good backup scheme.

### DIFF
--- a/docs/source/interface.rst
+++ b/docs/source/interface.rst
@@ -181,6 +181,22 @@ The response is **OK** with ``application/json`` content-type response body like
 
   { }
 
+``PATCH /storage-plugins/privatestorageio-zkapauthz-v1/unblinded-token``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This endpoint allows an external agent to discard unblinded tokens from the node's database.
+The unblinded tokens to trim are indicated by supplying a single unblinded token value.
+All unblinded tokens that appear before the given unblinded token in spend order will be discarded.
+
+The request body must be ``application/json`` encoded and contain an object like::
+
+  { "first-unspent": <unblinded token string>
+  }
+
+The response is **OK** with ``application/json`` content-type response body like::
+
+  { }
+
 ``POST /storage-plugins/privatestorageio-zkapauthz-v1/calculate-price``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/src/_zkapauthorizer/model.py
+++ b/src/_zkapauthorizer/model.py
@@ -555,6 +555,34 @@ class VoucherStore(object):
         )
 
     @with_cursor
+    def discard_unblinded_tokens_before(self, cursor, unblinded_token):
+        """
+        Get rid of all unblinded tokens that appear before the given token in
+        spend order.
+
+        :param UnblindedToken unblinded_token: The token that defines the
+            discard point.  All unblinded tokens that appear before this token
+            in spend order are discarded.
+
+        :return: ``None``
+        """
+        cursor.execute(
+            """
+            SELECT [rowid] FROM [unblinded-tokens] WHERE [token] = ?
+            """,
+            (unblinded_token.unblinded_token,),
+        )
+        (rowid,) = cursor.fetchone()
+        cursor.execute(
+            """
+            DELETE FROM [unblinded-tokens]
+            WHERE [rowid] < ?
+            """,
+            (rowid,),
+        )
+        cursor.execute("SELECT changes()")
+
+    @with_cursor
     def discard_unblinded_tokens(self, cursor, unblinded_tokens):
         """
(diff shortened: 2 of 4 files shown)


exarkun:
Here's the other part @crwood

tomprince (changes requested):
I think the code here looks fine, for what it is aiming to do, but as mentioned in https://github.com/PrivateStorageio/ZKAPAuthorizer/pull/237#issuecomment-946142191, I think we should preserve the discarded tokens, rather that delete them outright.

---

Also, I notice that the database doesn't track any kind of notion of time, for any of the tables. Thinking from our side, thinking about about what we might want to do if we do need to recover some incorrectly discarded or invalidated tokens, I could imagine wanting some notion of when the tokens were invalidated or discarded (though I guess when discarded isn't particularly meaningful). That said, adding dates is definitely out-of-scope for this PR. (I was thinking that dates might also be useful if we want to provide someway to refresh tokens after a key-rotation on our part, but I guess an associated public-key would be more useful for that)

exarkun:
Based on @tomprince 's excellent points above, this line of development doesn't seem worth pursuing.  We'll come up with a different plan to support improved durability of the internal state.

PR #242, opened 2021-11-02
Run black and isort over the codebase.

tomprince:
The only manual source changes are to `src/_zkapauthorizer/api.py`, which is in the final commit.

I will follow up with a commit that runs flake8, since that requires manual changes.

### DIFF
--- a/src/_zkapauthorizer/tests/test_storage_protocol.py
+++ b/src/_zkapauthorizer/tests/test_storage_protocol.py
@@ -39,9 +39,9 @@ from testtools.twistedsupport import (
     succeeded,
     failed,
 )
+# I'd rather use https://twistedmatrix.com/trac/ticket/8900 but efforts
+# there appear to have stalled.
 from testtools.twistedsupport._deferred import (
-    # I'd rather use https://twistedmatrix.com/trac/ticket/8900 but efforts
-    # there appear to have stalled.
     extract_result,
 )
 
@@ -94,8 +94,7 @@ from .strategies import (
     sizes,
     slot_test_and_write_vectors_for_shares,
     posix_timestamps,
-    # Not really a strategy...
-    bytes_for_share,
+    bytes_for_share,  # Not really a strategy...
 )
 from .matchers import (
     matches_version_dictionary,
--- a/src/_zkapauthorizer/tests/test_storage_protocol.py
+++ b/src/_zkapauthorizer/tests/test_storage_protocol.py
@@ -39,6 +39,7 @@ from testtools.twistedsupport import (
     succeeded,
     failed,
 )
+
 # I'd rather use https://twistedmatrix.com/trac/ticket/8900 but efforts
 # there appear to have stalled.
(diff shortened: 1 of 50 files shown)


src/_zkapauthorizer/tests/test_storage_protocol.py
```diff
@@ -16,128 +16,79 @@
...
 )
-from testtools.twistedsupport import (
-    succeeded,
-    failed,
-)
-from testtools.twistedsupport._deferred import (
-    # I'd rather use https://twistedmatrix.com/trac/ticket/8900 but efforts
-    # there appear to have stalled.
+from testtools.twistedsupport import failed, succeeded
+from testtools.twistedsupport._deferred import (  # I'd rather use https://twistedmatrix.com/trac/ticket/8900 but efforts; there appear to have stalled.
```
exarkun, line 39: will black let us put this comment on the line above instead, at least?

src/_zkapauthorizer/tests/test_storage_protocol.py
```diff
@@ -16,128 +16,79 @@
...
+    cleanup_storage_server,
+    get_passes,
+    pass_factory,
+    privacypass_passes,
+    whitebox_write_sparse_share,
+    write_toy_shares,
 )
-from ..foolscap import (
-    ShareStat,
+from .strategies import (  # Not really a strategy...
```
exarkun, line 72: arguably the meaning of this comment has been lost by the change

exarkun (approved):
Thanks.  I had some WTF moments reading the new formatting.  However, hooray consistency and tool enforcement all that.

There are a couple changes to comment formatting that I hope can be fixed but otherwise this seems landable.

Please fix the points noted inline to the extent possible and then merge.

src/_zkapauthorizer/tests/test_storage_protocol.py
tomprince: I've made this change in the flake8 PR. I'll note that it is actually isort that makes this change; I searched for a while to see if there was a way to disable moving comments, unsuccessfully.

src/_zkapauthorizer/_version.py
```diff
@@ -1,4 +1,3 @@
-
```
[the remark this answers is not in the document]
tomprince: I've updated the configuration to skip this file (and reverted the change to it).

src/_zkapauthorizer/tests/test_storage_protocol.py
tomprince: (I've actually moved it above in this PR now).

src/_zkapauthorizer/tests/test_storage_protocol.py
tomprince: After some experimentation, I've found that trailing comments on an import line are preserved, so I've moved this to the appropriate place. I also filed https://github.com/PyCQA/isort/issues/1843 to see if this behavior can be better documented.

PR #244, opened 2021-11-08
Change implementer_only to implementer.

### LINKED ISSUE #9: Figure out the weirdness with adaption relating to SecureAccessTokenAuthorizerStorageServer
`_storage_server.py` ends with

```
# I don't understand why this is required.                                                                                                                                                                                                                                     
# SecureAccessTokenAuthorizerStorageServer is-a Referenceable.  It seems like                                                                                                                                                                                                  
# the built in adapter should take care of this case.                                                                                                                                                                                                                          
from twisted.python.components import (
    registerAdapter,
)
from foolscap.referenceable import (
    ReferenceableSlicer,
)
from foolscap.ipb import (
    ISlicer,
)
registerAdapter(ReferenceableSlicer, SecureAccessTokenAuthorizerStorageServer, ISlicer)
```

As the comment says, it's not clear to me why this is necessary.  Figure it out and fix whatever the issue is so this code can go away or explain why it needs to remain.

tomprince:
This removes the need to call `registerAdapter` on
`ZKAPAuthorizerStorageServer`. It looks like that was needed because the
interfaces for it were defined by `implementer_only` which explicitly ignored
interfaces defined by parent classes. Looking at the history, I assume the use
of `implementer_only` was to avoid implementing `RIStorageServer` via
`proxyForInterface`. Given that the later is no longer used, `implementer_only`
is no longer necessary.

This is extracted from #243.

Fixes #9

### DIFF
--- a/src/_zkapauthorizer/_storage_server.py
+++ b/src/_zkapauthorizer/_storage_server.py
@@ -38,11 +38,10 @@ from attr.validators import instance_of, provides
 from challenge_bypass_ristretto import SigningKey, TokenPreimage, VerificationSignature
 from eliot import start_action
 from foolscap.api import Referenceable
-from foolscap.ipb import IReferenceable, IRemotelyCallable
 from twisted.internet.defer import Deferred
 from twisted.internet.interfaces import IReactorTime
 from twisted.python.reflect import namedAny
-from zope.interface import implementer_only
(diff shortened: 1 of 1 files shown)


exarkun:
Thanks!

exarkun (approved):

PR #247, opened 2021-11-11
Use hypothesis setup_example hook

### LINKED ISSUE #246: Hypothesis `setup_example` hook could be used to remove a lot of duplication in the test suite
A lot of tests start off with some cleanup to make them work well under Hypothesis, where a test method is run repeatedly with different arguments.

Hypothesis provides the `setup_example` hook which runs before each such "example" and could be used to do this cleanup in a single place instead of in every test.

exarkun:
Fixes #246


exarkun:
Looks good to *me* :wink:

PR #248, opened 2021-11-12
Turn off implicit lease renewal for allocate_buckets

### LINKED ISSUE #219: `allocate_buckets` spending doesn't account for renewed leases.
Currently, `allocate_buckets` only pays/charges for newly allocated buckets. However, in addition to allocating new buckets, any existing buckets for that storage index are renewed.

exarkun:
Fixes #219

### DIFF
--- a/src/_zkapauthorizer/_storage_server.py
+++ b/src/_zkapauthorizer/_storage_server.py
@@ -179,13 +179,6 @@ class ZKAPAuthorizerStorageServer(Referenceable):
         # still add leases to the newly uploaded shares.
         self._original.set_implicit_bucket_lease_renewal(False)
 
-        # Similarly, wrapped ``slot_testv_and_readv_and_writev_message``
-        # renews leases on all shares that are being modified.  Turn that
-        # behavior off.  This means we have to take responsibility for
-        # creating the initial lease on shares when they are created (and we
-        # do in our wrapper for ``slot_testv_and_readv_and_writev_message``).
-        self._original.set_implicit_slot_lease_renewal(False)
-
     def remote_get_version(self):
         """
         Pass-through without pass check to allow clients to learn about our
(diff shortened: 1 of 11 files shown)


src/_zkapauthorizer/_storage_server.py
```diff
@@ -173,6 +173,12 @@ class ZKAPAuthorizerStorageServer(Referenceable):
         default=attr.Factory(partial(namedAny, "twisted.internet.reactor")),
     )
 
+    def __attrs_post_init__(self):
+        # Avoid the default StorageServer ``allocate_buckets`` behavior of
+        # renewing leases on all existing shares in the same bucket.  It will
+        # still add leases to the newly uploaded shares.
+        self._original.set_implicit_bucket_lease_renewal(False)
```
tomprince, line 180: Per https://github.com/tahoe-lafs/tahoe-lafs/pull/1160#issuecomment-967446029, I don't think that setting this here makes sense (as opposed to passing in the value to the internal call).

src/_zkapauthorizer/_storage_server.py
```diff
@@ -360,6 +366,7 @@ def _slot_testv_and_readv_and_writev(
             if required_new_passes > len(validation.valid):
                 validation.raise_for(required_new_passes)
 
+        self._original.set_implicit_slot_lease_renewal(renew_leases)
```
tomprince, line 369: I was curious why this switched to being a instance attribute on the wrapped server, but I see you change to set this globally in #249. As mentioned in the tahoe PR this depends on, I'm not sure that having an instance attribute, as opposed to a parameter is the best interface, though.

tomprince (approved):
This looks fine, given the current state of https://github.com/tahoe-lafs/tahoe-lafs/pull/1160, but as mentioned in https://github.com/tahoe-lafs/tahoe-lafs/pull/1160#issuecomment-967446029 I don't know if the interface there is correct.

PR #249, opened 2021-11-12
Improve lease management for mutables

### LINKED ISSUE #221: Leases for mutable files are handled incorrectly in `slot_testv_and_readv_and_writev`
There are a couple of things that can happen:
- the writev could indicate that a new sharefile should be created on this server. In this case, the sharefile will not have a lease applied to it, so the file could be collected immediately.
- the first share found doesn't have a lease (probably from the first case). In this case, we will treat things as if no sharefiles have a lease[1], so pay for the entire existing size of the file, and extend the lease, losing the value of any existing leases.
- Otherwise, we pay for the difference in size between the current and new sizes, but don't change the lease time, losing the paid value beyond the current lease time.

[1] `StorageServer.get_slot_leases` (called by `has_active_leases`) returns the lease from the first sharefile it finds. It looks like that function is not called in tahoe at all.

exarkun:
Fixes #221 

The old behavior:

* If an arbitrarily selected existing share in a slot where a write is happening has an unexpired lease, no share written by the operation will have a lease added or renewed (including new shares, which therefore end up with no lease at all).
* If existing shares have unexpired leases then the client counts the excess value of those leases as credit towards the operation it is performing but the server only counts them if they are being written to and so clients under-fund certain write operations.

The new behavior:

* A mutable write will add leases to or renew leases on exactly the set of shares which have no unexpired leases and are also being written to.
* The server takes all shares in the slot into account when determining the number of additional passes the client must spend to complete the operation so that it comes to the same conclusion as the client about how many passes are required.


tomprince:
> * A mutable write will add leases to or renew leases on exactly the set of shares which have no unexpired leases and are also being written to.
> 
> * The server takes all shares in the slot into account when determining the number of additional passes the client must spend to complete the operation so that it comes to the same conclusion as the client about how many passes are required.

It would probably be useful to write down how the number of passes required here differs from the amount that would be required if passes were infinitely divisible, and usable for different shapes of storage time.

- If there are existing leases on sharefiles being written to, we charge for any increase in size for one lease period but we *don't* update the lease for that file. The total we will have charged is
  ```
  remaining_lease * existing_size + lease_period * size_increase
  ```
  but we provide
  ```
  remaining_lease * (existing_size + size_increase)
  ```
  in terms of storage time, so the client is being overcharged
  ```
  (lease_period - remaining_lease) * size_increase
  ```

src/_zkapauthorizer/_storage_server.py
```diff
@@ -695,3 +701,100 @@ def get_stat(sharepath):
...
+    """
+    Determine whether the given share on the given server has an unexpired
+    lease or not.
+
+    :return: ``True`` if it has at least one unexpired lease, ``False``
+        otherwise.
+    """
+    sharepath = get_share_path(storage_server, storage_index, sharenum)
+    share = get_share_file(sharepath.path)
+    return any(lease.get_expiration_time() > now for lease in share.get_leases())
```
tomprince, line 757: I *think* this should probably have an `if lease.renew_secret == renew_secret` here, but that will require changes elsewhere.

src/_zkapauthorizer/_storage_server.py
```diff
@@ -695,3 +701,100 @@ def get_stat(sharepath):
...
+    """
+    # Find the current size of shares being written.
+    current_sizes = dict(
+        get_share_sizes(
+            storage_server,
+            storage_index,
+            # Consider the size of *all* shares even if they're not being
+            # written.  If they have an unexpired lease then we can apply some
+            # or all of the remainder of the value of that lease towards this
+            # operation.
```
tomprince, line 773: I'm not sure how meaningful this comment is. I was going to say that sharefiles that aren't touched are going to have the same size before and after in `get_required_new_passes_for_mutable_write`, but I guess that size is taken into account when `required_passes` is rounding the number of passes required. However, I think that can only ever change the number of passes required by ±1, but that doesn't really seem like applying the remaining value of those leases to anything we charge here.

---

As an aside, it does feel a little weird that existing share can have that effect. I guess a client can always zero out the size of any shares it doesn't care about.

src/_zkapauthorizer/_storage_server.py
```diff
@@ -292,12 +307,25 @@ def remote_slot_testv_and_readv_and_writev(
...
-            data in already-allocated storage.  These cases may not be the
-            same from the perspective of pass validation.
+        * If shares are created then a lease is added to them.
+          Passes are required for the full size of the share.
+
+        * If shares without unexpired leases are written then a lease is added to them.
+          Passes are required for the full size of the shares after the write.
+
+        * If shares with unexpired leases are made larger.
+          Passes are required for the difference in price between the old and new size.
```
tomprince, line 328: As noted in the PR comment, a client will lose some value here (inversely proportional to how much of a remaining lease period there is on those shares).

src/_zkapauthorizer/tests/test_storage_protocol.py
```diff
@@ -1115,3 +1108,16 @@ def write_vector_to_read_vector(write_vector):
...
     """
     return (write_vector[0], len(write_vector[1]))
+
+
+def get_lease_grant_times(storage_server, storage_index):
+    """
+    Get the grant times for all of the leases for all of the shares at the
+    given storage index.
+    """
+    shares = storage_server._get_bucket_shares(storage_index)
```
tomprince, line 1118: I'm curious about the use of this here vs. `get_all_share_paths` in `_storage_server.py`, and the new `get_share_path`. It seems like there are quite a few different places that track the on-disk layout.

tomprince (changes requested):
This looks good, and an improvement on the behaviour of the existing code. There are a couple of comments about things we should consider in the future (whether to restrict which leases we look at, and the difference between what is charged vs value provided) that don't need to be addressed here. There are also some comments that I think are unclear or incorrect, that I think should be changed before this lands; the behaviour feels subtle enough that I'd like to review those changes.

src/_zkapauthorizer/_storage_server.py
```diff
@@ -323,56 +351,47 @@ def _slot_testv_and_readv_and_writev(
...
+        result = self._original.remote_slot_testv_and_readv_and_writev(
             storage_index,
             secrets,
             tw_vectors,
             r_vector,
-            renew_leases=renew_leases,
         )
 
+        # Add the lease that we charged the client for: leases on all written
+        # shares without an unexpired lease.
```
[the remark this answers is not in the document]
exarkun: Addressed in b8851317bc684e1922633dd55e7e3a21fd11149f I believe

src/_zkapauthorizer/_storage_server.py
exarkun: It's possible.  I think it will depend on how we want billing and sharing to interact, though.  Since we're basically just not doing sharing yet I think any interaction (eg, the current behavior) is fine for now.

src/_zkapauthorizer/_storage_server.py
exarkun: I expanded on this comment in 6789754d and I think I gave an example where the required number of passes can change by more than 1... but at a higher level, I agree this is weird behavior.  It's also a bit of a tangent to the issue this PR is meant to address.  I think that what happened is when I made the Hypothesis-based test for mutables more comprehensive, it uncovered another existing bug where the client and server disagreed on the price calculation.  The client was already counting all existing shares but the server was not.

It probably would have been better to split this out into a separate ticket/PR but I forgot to.  I still could, but maybe since you already spent the time reading the diff it's not particularly worthwhile now.

I think the behavior implemented here is not ideal but it is better than before (because it will not cause spending errors due to client/server disagreement in this case anymore).

src/_zkapauthorizer/_storage_server.py
exarkun: Added a note to this docstring in 9abf3a50

hacklschorsch (approved):
LGTM, but I can't say I dug in deep enough to understood every detail.

src/_zkapauthorizer/tests/test_storage_protocol.py
exarkun: It's a mess.  I wouldn't mind having some time to do some cleanup, on the Tahoe-LAFS side as well as the ZKAPAuthorizer side.

Some of the inconsistency comes from changing my mind over time about how much work it is worth doing to avoid depending on yet another Tahoe-LAFS Python API by re-implementing something in ZKAPAuthorizer.  It would be nice to have a chance to make a pass over the codebase to bring it all into consistency with a single stance on this subject.

src/_zkapauthorizer/_storage_server.py
exarkun: > I don't think your example works as you describe, since we don't ever take the sum of quantized values.

Okay.  I think you're right.  Eventually I decided to change the logic to go the other way.  Now the client and server both agree to ignore existing shares.

tomprince (approved):

PR #250, opened 2021-11-12
Update to a version of tahoe that includes openmetrics support.

tomprince:
Fixes https://whetstone.privatestorage.io/privatestorage/PrivateStorageio/-/issues/90.

It turns out that the upstream change was only on the client side, so this changes the value as soon as we receive it on the client side.

### DIFF
--- a/src/_zkapauthorizer/_storage_client.py
+++ b/src/_zkapauthorizer/_storage_client.py
@@ -424,6 +424,23 @@ class ZKAPAuthorizerStorageClient(object):
         # Read operations are free.
         num_passes = 0
 
+        # Convert tw_vectors from the new internal format to the wire format.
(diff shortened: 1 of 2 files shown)


src/_zkapauthorizer/_storage_client.py
```diff
@@ -424,6 +424,23 @@ def slot_testv_and_readv_and_writev(
         # Read operations are free.
         num_passes = 0
 
+        # Convert tw_verctors from the new internal format to the wire format.
```
exarkun, line 427: typo here, `tw_verctors`

exarkun (approved):
Thanks.  Looks good.  One trivial point inline.  Please address and merge.

PR #252, opened 2021-11-16
Better lease renewal decisions

### LINKED ISSUE #223: Correct lease expiration times are not necessarily used for lease maintenance spending
Lease renewal decisions, as made by `_zkapauthorizer/lease_maintenance.py` and `renew_leases_on_server` in particular, are based on the results of `stat_shares` as implemented by our server-side storage server wrapper.

The server looks up lease expiration information in `_zkapauthorizer/_storage_server.py` (`get_lease_expiration`).  This function returns the expiration time of the *first* lease it comes across for the indicated storage index.

* How is "first" defined?  Is it stable?
* What if the "first" lease does not belong to the particular client inspecting the lease state?
* What if the "first" lease does not belong to the particular client inspecting the lease state and it is *expired* already?

exarkun:
Fixes #223

### DIFF
--- a/src/_zkapauthorizer/tests/matchers.py
+++ b/src/_zkapauthorizer/tests/matchers.py
@@ -135,12 +135,12 @@ def leases_current(relevant_storage_indexes, now, min_lease_remaining):
     servers for which the leases on the given storage indexes do not expire
     before ``min_lease_remaining``.
     """
+
     def get_relevant_stats(storage_server):
         for (storage_index, shares) in storage_server.buckets.items():
             if storage_index in relevant_storage_indexes:
                 for (sharenum, stat) in shares.items():
                     yield stat
-
 
     return AfterPreprocessing(
         # Get share stats for storage indexes we should have
--- a/src/_zkapauthorizer/tests/matchers.py
+++ b/src/_zkapauthorizer/tests/matchers.py
@@ -140,8 +140,9 @@ def leases_current(relevant_storage_indexes, now, min_lease_remaining):
         # visited and maintained.
         lambda storage_server: list(
             stat
-            for (storage_index, stat) in storage_server.buckets.items()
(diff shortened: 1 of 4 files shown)


src/_zkapauthorizer/tests/matchers.py
```diff
@@ -140,8 +140,9 @@ def leases_current(relevant_storage_indexes, now, min_lease_remaining):
         # visited and maintained.
         lambda storage_server: list(
             stat
-            for (storage_index, stat) in storage_server.buckets.items()
+            for (storage_index, shares) in storage_server.buckets.items()
             if storage_index in relevant_storage_indexes
+            for (sharenum, stat) in shares.items()
```
hacklschorsch, line 145: I find this difficult to understand.
It is not a list comprehension, but something similar;
it lives in a list constructor.
It has no indentation.

list(stat for (tuple) in (list) if (relevant) for (shares) ... ❓

Can you give me a pointer so I can RTFM?

hacklschorsch (approved):
LGTM!

I find the code itself good to understand and _pythonic_;
I can't make much sense of the tests.

Maybe it's the combination of _testtools_ with _hypothesis_.
I fail to keep up with the many indirections.

src/_zkapauthorizer/tests/matchers.py
exarkun: Yea this syntax sucks.  I'll rewrite it using a for loop statement instead.  For your reference, there's https://docs.python.org/3/tutorial/datastructures.html#list-comprehensions but I don't know if it makes any sense to someone who doesn't already know everything it is trying to say.

This *is* a list comprehension - with two loops in it.  The outer loop is the first `for` and has a filtering condition that applies to the inner loop which is the second `for`.  The elements that end up in the list are the values of the expression at the beginning, as usual, evaluated once for each iteration of the inner loop within each iteration of the outer loop.

PR #253, opened 2021-11-16
Make generating cross-plaform pins easier.

tomprince:
This add colorama as a direct dependency. Currently, it a transitive dependency only on windows. When using `pip-compile` (from pip-tools), to pin dependencies, it only supports generating dependencies for the current platform. However, there are only two dependencies that vary between platforms, colorama and pywin32. pywin32 can't be installed on non-windows platform and has dropped python 2 support, so is relatively easy to pin. colorama is cross-platform, so including it here is a tiny bit wasteful, but makes pinning it trivial.

This is part of implementing https://whetstone.privatestorage.io/privatestorage/privatestoragedesktop/-/issues/603 for gridsync (see https://github.com/gridsync/gridsync/pull/400).

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -49,6 +49,13 @@ install_requires =
     treq
     pyutil
     prometheus-client
+    # Include colorama as a dependency to help pip-compile deal with multiple
+    # platforms.  In particular, tqdm depends on colorama only on Windows. By
+    # including it here, pip-compile will generate hashes (and install it) on
+    # all platforms. colorama and pywin32 are our only depdencies that are only
+    # required on some platforms; we can't include pywin32 here as it does not
+    # install cross-platform.
+    colorama
 
 [options.extras_require]
 test = coverage; fixtures; testtools; hypothesis


crwood (approved):
While it is indeed wasteful to have to declare dependencies for platforms that don't actually require them like this (and, especially, in this particular case, since, to my knowledge, we're not actually using `tqdm` directly anywhere and are only inheriting it to satisfy the Magic-Wormhole CLI usage -- which we _also_ don't use), given the current state of the tooling available in the python packaging ecosystem, this appears to the most straightforward solution. (Doing this the "Right" way, it seems, would require generating pins on each platform individually -- _and_ across multiple python interpreter versions -- and then merging the results but, to me, the added complexity of doing that doesn't seem worth it here given that we're only talking about one sub-sub-dependency that existing codepaths don't even touch...).

So, LGTM. :)

(Also, I like that, after this is merged, I won't have to continue to manually preserve the `colorama` line in [this file](https://github.com/gridsync/gridsync/blob/a01001e016d099c7d27547ae5b15e7b903249530/requirements/tahoe-lafs.in) -- so that's a bonus.)

PR #255, opened 2021-11-18
Add configuration for lease maintenance behavior

### LINKED ISSUE #169: Accept configuration controlling the lease crawler scheduling and other behavior
To maintain stored data, ZKAPAuthorizer periodically crawls capabilities reachable from the rootcap (where directory nodes provide additional edges in the graph to crawl).

There are certain magic numbers involved in this process:

  - How frequently does it crawl?
  - How close must a lease be to expiration for the crawler to renew it?

At some point in the future there may also be a magic number for:

  - How far into the future should leases be renewed?

Rather than have ZKAPAuthorizer make these decisions for all users, it should accept these as inputs which can then vary on a per-grid (perhaps even per-user, in some cases) basis.

ZKAPAuthorizer already takes some configuration from a custom section in tahoe.cfg so this seems like a reasonable place for this new configuration as well.  Therefore, the configuration schema should be extended:

```
[storageclient.plugins.privatestorageio-zkapauthz-v1]
...
lease.crawl-interval.mean = <time string>
lease.crawl-interval.range = <time string>
lease.min-time-remaining = <time string>
```

`lease.crawl-interval.mean` and map directly on to the `interval_mean` and `interval_range` parameters of `_zkapauthorizer.lease_maintenance.lease_maintenance_service`.  `lease.min-time-remaining` maps directly on to the `min_lease_remaining` parameter of `maintain_leases_from_root`.

exarkun:
Fixes #169 

~~Probably done but I'm going to integrate it w/ GridSync before saying so for sure.~~

Per discussion elsewhere, GridSync also/instead needs a different integration point for answering the question it has.  So this is probably fine as-is (and maybe also not useful in the near term - but still maybe useful in the long term, and at least not harmful right now).

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -89,7 +89,7 @@ The value is an integer number of seconds.
 For example to make all intervals fall within a 7 day period::
 
   [storageclient.plugins.privatestorageio-zkapauthz-v1]
-  lease.crawl-interval.mean = 302400
+  lease.crawl-interval.range = 302400
 
 
 lease.min-time-remaining
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -66,6 +66,42 @@ When tokens are received from an issuer during redemption,
 these are the only public keys which will satisfy the redeemer and cause the tokens to be made available to the client to be spent.
 Tokens received with any other public key will be sequestered and will *not* be spent until some further action is taken.
 
+lease.crawl-interval.mean
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This item controls the frequency at which the lease maintenance crawler runs.
+The lease maintenance crawler visits all shares and renews their leases if necessary.
+The crawler will run at random intervals.
+The client will try to make the average (mean) interval between runs equal to this setting.
+The value is an ISO8601 duration string.
+For example to run on average every 26 days::
+
+  [storageclient.plugins.privatestorageio-zkapauthz-v1]
+  lease.crawl-interval.mean = P26D
+
+lease.crawl-interval.range
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This item also controls the frequency of lease maintenance crawler runs.
+The random intervals between runs have a uniform distribution with this item's value as its range.
+The value is an ISO8601 duration string.
+For example to make all intervals fall within a 7 day period::
+
+  [storageclient.plugins.privatestorageio-zkapauthz-v1]
+  lease.crawl-interval.mean = P3DT12H
+
+lease.min-time-remaining
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+This item controls the lease renewal behavior of the lease maintenance crawler.
+It specifies an amount of time left on a lease.
+If the crawler encounters a lease with less time left than this then it will renew the lease.
+The value is an ISO8601 duration string.
+For example to renew leases on all shares which will expire in less than one week::
+
+  [storageclient.plugins.privatestorageio-zkapauthz-v1]
+  lease.min-time-remaining = P7D
+
 Server
 ------
 
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -73,22 +73,24 @@ This item controls the frequency at which the lease maintenance crawler runs.
 The lease maintenance crawler visits all shares and renews their leases if necessary.
 The crawler will run at random intervals.
 The client will try to make the average (mean) interval between runs equal to this setting.
-The value is an ISO8601 duration string.
+The value is an integer number of seconds.
 For example to run on average every 26 days::
 
   [storageclient.plugins.privatestorageio-zkapauthz-v1]
-  lease.crawl-interval.mean = P26D
+  lease.crawl-interval.mean = 2246400
+
 
 lease.crawl-interval.range
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 This item also controls the frequency of lease maintenance crawler runs.
 The random intervals between runs have a uniform distribution with this item's value as its range.
-The value is an ISO8601 duration string.
+The value is an integer number of seconds.
 For example to make all intervals fall within a 7 day period::
 
   [storageclient.plugins.privatestorageio-zkapauthz-v1]
-  lease.crawl-interval.mean = P3DT12H
+  lease.crawl-interval.mean = 302400
+
 
 lease.min-time-remaining
 ~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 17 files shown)


docs/source/configuration.rst
```diff
@@ -66,6 +66,42 @@ When tokens are received from an issuer during redemption,
...
 Tokens received with any other public key will be sequestered and will *not* be spent until some further action is taken.
 
+lease.crawl-interval.mean
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This item controls the frequency at which the lease maintenance crawler runs.
+The lease maintenance crawler visits all shares and renews their leases if necessary.
+The crawler will run at random intervals.
+The client will try to make the average (mean) interval between runs equal to this setting.
+The value is an ISO8601 duration string.
```
tomprince, line 76: I don't think I've ever encountered this format, so I wondered a bit about the use of it here. Looking at the examples here and the discussion [on wikipedia](https://en.wikipedia.org/wiki/ISO_8601#Durations), the format looks reasonable, and readable (once you ignore the leading `P`).

It does have an alternative format (which is like `P0000-00-26T00:00:00`), which is looks like isodate has bunch of bugs in parsing (and is unmaintained). There is also isoduartion (which was inspired by iso date), but it also has a bunch of bugs in parsing the alternative format.

I've gone back-and-forth a bit on whether it makes sense to use this format:
- My first thought was that this was a weird thing, and maybe we shouldn't use it. I was going to point out that [RFC3339](https://datatracker.ietf.org/doc/html/rfc3339) is the format used in practice for dates which is a restricted profile of ISO 8601, though people often refer to that format as ISO 8601; and that the other bits of ISO 8601 are not commonly encountered.
- After thinking about it a bit more, and looking at the wikipedia page for the spec, I decided that it seemed like a reasonable format; it the regular form is relatively unambiguous[1] and mostly human readable. Thus, while a bit odd, it seemed like a reasonable choice.
- However, when I came back to look at it again, I got curious about the alternate format, and went to look at how the parsing for it was implemented. Both isodate and isoduration defer to a date parse for the alternative format, which is incorrect[2]. Given the state of parsers, I'm leaning away from using it; The appeal of using a standardized format like this is to re-use existing code, but the quality of the implementations I can find don't hold up.

[1] Though as pointed out on the wikipedia page, `P1M` could be ambiguous.
[2] It at least accepts too much, since it accepts any date format (isodate uses its own, isoduration uses arrow's), which both accept more than what is defined for ISO 8061. I think they may accept to little (though the copies of ISO 8061 I could find aren't clear on this), as they don't let you have have 0 for any of the components.

exarkun: Did you have any ideas for an alternative format to use?

tomprince: Given that we aren't exposing this value to users, we could probably get away with something simple like number of seconds or days. Slightly more complex, we could allow suffixes for days/hours/minutes/seconds.

exarkun: Oh yea.  Number of seconds seems fine.  Thanks.

docs/source/configuration.rst
```diff
@@ -66,6 +66,42 @@ When tokens are received from an issuer during redemption,
...
+lease.crawl-interval.range
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This item also controls the frequency of lease maintenance crawler runs.
+The random intervals between runs have a uniform distribution with this item's value as its range.
+The value is an ISO8601 duration string.
+For example to make all intervals fall within a 7 day period::
+
+  [storageclient.plugins.privatestorageio-zkapauthz-v1]
+  lease.crawl-interval.mean = P3DT12H
```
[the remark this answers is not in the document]
exarkun: I suppose.  Turns out all of the arithmetic averages should be the same for the distribution we use.  "mean" isn't *wrong* here though is it?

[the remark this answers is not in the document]
tomprince: There isn't quite enough context here to make it clear, but this is in the `range` section. It looks like I also made a mistake when writing the comment about that, which makes the suggestion as-is confusing. It should be replaced with `range` not `average`.

docs/source/configuration.rst
```diff
@@ -66,6 +66,44 @@ When tokens are received from an issuer during redemption,
...
+lease.crawl-interval.range
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This item also controls the frequency of lease maintenance crawler runs.
+The random intervals between runs have a uniform distribution with this item's value as its range.
+The value is an integer number of seconds.
+For example to make all intervals fall within a 7 day period::
+
+  [storageclient.plugins.privatestorageio-zkapauthz-v1]
+  lease.crawl-interval.mean = 302400
```
tomprince, line 92:
```suggestion
  lease.crawl-interval.range = 302400
```

docs/source/configuration.rst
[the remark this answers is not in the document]
exarkun: ahh that makes more sense, thanks for the clarification.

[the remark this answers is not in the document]
tomprince: It took me a bit of time to figure it out too. I spent some time trying to figure out why I would have made the suggestion, which I recalled being very obvious and not worthy of comment, and trying to figure out why I would have suggested changing `mean` to `average`. :upside_down_face: Then when I looked at the diff in context, I finally realized it wasn't in the section I assumed it was.

src/_zkapauthorizer/config.py
```diff
@@ -0,0 +1,98 @@
...
+    Tahoe-LAFS config.
+
+    :param cfg: The Tahoe-LAFS config object to consult.
+    :param option: The name of the option to read.
+
+    :return: ``None`` if the option is missing, otherwise the parsed duration
+        as a ``timedelta``.
+    """
+    # type: (_Config, str) -> Optional[timedelta]
+    section_name = u"storageclient.plugins.privatestorageio-zkapauthz-v1"
```
tomprince, line 90: This feels like a weird place for this constant. I'd prefer if this were passed in from `_plugin`. (If we had infinite time, I'd likely suggest having an object wrapping the node config, that reads from a particular section, but it probably isn't worth the effort here yet).

src/_zkapauthorizer/config.py
```diff
@@ -0,0 +1,98 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Helpers for reading values from the Tahoe-LAFS configuration.
+"""
```
tomprince, line 17: I wonder if we want to have separate modules for client and server config. Other than `empty_config` (which I guess is a test fixture), all the stuff here is client specific.

tomprince (approved):
Looks good. It would be nice if the config section to read from was not hard-coded in a leaf function, but that isn't blocking.

src/_zkapauthorizer/config.py
exarkun: Yea.  Filed https://github.com/PrivateStorageio/ZKAPAuthorizer/issues/262

src/_zkapauthorizer/config.py
exarkun: Could be, yea.  Since this project started it has become clear that two different plugins may have made more sense than a single plugin with such a tightly coupled implementation.  We could probably try to keep the pieces as separate as possible until a more definitive decoupling happens.  I'll keep it in mind for future development.

PR #257, opened 2021-12-03
Fold `PassValidationTests._test_lease_operation_fails_without_passes` into its caller.

tomprince:
It was originally factored out since there were both `add_lease` and `renew_lease methods`.
However, since tahoe-lafs 1.16, the client no longer defines `renew_lease`, and we removed it
from our protocol in #231. Given that there is now only one caller, we can remove the
factoring.

I ran into this, while working out how to add tests for tracking spending.


exarkun (approved):
Nice, thanks.

PR #258, opened 2021-12-03
Fix pyflakes invocation in nix.

tomprince:
It turns our that `pyflakes` with no arguments expects input
on stdin. In CI, this is likely connected to `/dev/null` so it
passes.

Also fix the errors that pyflakes is reporting.

### DIFF
--- a/src/_zkapauthorizer/_storage_server.py
+++ b/src/_zkapauthorizer/_storage_server.py
@@ -31,11 +31,10 @@ from os.path import join
 from struct import calcsize, unpack
 
 import attr
-from allmydata.interfaces import RIStorageServer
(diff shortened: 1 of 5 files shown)


src/_zkapauthorizer/_storage_server.py
```diff
@@ -60,6 +59,11 @@
     slot_testv_and_readv_and_writev_message,
 )
 
+try:
+    from typing import Dict, List, Optional
+except ImportError:
+    pass
```
tomprince, line 65: I copied this idiom from tahoe-lafs.

exarkun (approved):
Ah, good catch, thanks.

PR #263, opened 2021-12-06
Convert `Pass`, `UnblindedToken`, `RandomToken` and `Voucher` to use bytes internally.

tomprince:
As part of working on https://whetstone.privatestorage.io/privatestorage/PrivateStorageio/-/issues/84, I want to extract the passes being spent in `_ValidationResult.validate_passes`, to do that, I'd like to reuse the parsing in `_zkapauthorizer.model.Pass`. However, currently that only accepts `unicode`, but `_ValidationResult.validate_passes` has `bytes`. While I could convert to `unicode` to parse, looking at where `Pass` and `UnblindedToken`, it seems like it would be simpler to have them accept `bytes` instead.

I'm a little less sure regarding `Voucher`, I changed it partly to match the other model objects, and partly because I changed the attrs validator that was being used.

### DIFF
--- a/src/_zkapauthorizer/model.py
+++ b/src/_zkapauthorizer/model.py
@@ -247,7 +247,7 @@ class VoucherStore(object):
     @with_cursor
     def get(self, cursor, voucher):
         """
-        :param unicode voucher: The text value of a voucher to retrieve.
+        :param bytes voucher: The text value of a voucher to retrieve.
 
         :return Voucher: The voucher object that matches the given value.
         """
@@ -260,7 +260,7 @@ class VoucherStore(object):
             WHERE
                 [number] = ?
             """,
-            (voucher,),
+            (voucher.decode("ascii"),),
         )
         refs = cursor.fetchall()
         if len(refs) == 0:
@@ -274,7 +274,7 @@ class VoucherStore(object):
         existing) to the database.  If the (voucher, counter) pair is already
         present, do nothing.
 
-        :param unicode voucher: The text value of a voucher with which to
+        :param bytes voucher: The text value of a voucher with which to
             associate the tokens.
 
         :param int expected_tokens: The total number of tokens for which this
@@ -302,7 +302,7 @@ class VoucherStore(object):
             FROM [tokens]
             WHERE [voucher] = ? AND [counter] = ?
             """,
-            (voucher, counter),
+            (voucher.decode("ascii"), counter),
         )
         rows = cursor.fetchall()
         if len(rows) > 0:
@@ -312,26 +312,35 @@ class VoucherStore(object):
                 voucher=voucher,
                 counter=counter,
             )
-            tokens = list(RandomToken(token_value) for (token_value,) in rows)
+            tokens = list(
+                RandomToken(token_value.encode("ascii")) for (token_value,) in rows
+            )
         else:
             tokens = get_tokens()
             self._log.info(
                 "Persisting {count} random tokens for a voucher ({voucher}[{counter}]).",
                 count=len(tokens),
-                voucher=voucher,
+                voucher=voucher.decode("ascii"),
                 counter=counter,
             )
             cursor.execute(
                 """
                 INSERT OR IGNORE INTO [vouchers] ([number], [expected-tokens], [created]) VALUES (?, ?, ?)
                 """,
-                (voucher, expected_tokens, self.now()),
+                (voucher.decode("ascii"), expected_tokens, self.now()),
             )
             cursor.executemany(
                 """
                 INSERT INTO [tokens] ([voucher], [counter], [text]) VALUES (?, ?, ?)
                 """,
-                list((voucher, counter, token.token_value) for token in tokens),
+                list(
+                    (
+                        voucher.decode("ascii"),
+                        counter,
+                        token.token_value.decode("ascii"),
+                    )
+                    for token in tokens
+                ),
             )
         return tokens
 
@@ -385,7 +394,7 @@ class VoucherStore(object):
         """
         Store some unblinded tokens received from redemption of a voucher.
 
-        :param unicode voucher: The voucher associated with the unblinded
+        :param bytes voucher: The voucher associated with the unblinded
             tokens.  This voucher will be marked as redeemed to indicate it
             has fulfilled its purpose and has no further use for us.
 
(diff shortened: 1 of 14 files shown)


src/_zkapauthorizer/validators.py
```diff
@@ -22,10 +22,10 @@
 def is_base64_encoded(b64decode=b64decode):
     def validate_is_base64_encoded(inst, attr, value):
         try:
-            b64decode(value.encode("ascii"))
+            b64decode(value)
```
exarkun, line 25: Do you mind adding a docstring and/or type comment to `is_base64_encoded` that reflects this change?

src/_zkapauthorizer/tests/test_client_resource.py
```diff
@@ -493,7 +493,8 @@ def test_post(self, get_config, api_auth_token, voucher, unblinded_tokens):
             dumps(
                 {
                     u"unblinded-tokens": list(
-                        token.unblinded_token for token in unblinded_tokens
+                        token.unblinded_token.encode("ascii")
```
exarkun, line 496: `unblinded_tokens` is a list of `UnblindedToken` and `UnblindedToken.unblinded_token` is now `bytes` instead of `unicode` so this `encode` is unnecessary (though "safe" because it's a base64 string).  Right?

src/_zkapauthorizer/tests/storage_common.py
```diff
@@ -154,7 +154,7 @@ def get_passes(message, num_passes):
 
 def get_passes(message, count, signing_key):
     """
-    :param unicode message: Request-binding message for PrivacyPass.
+    :param bytes message: Request-binding message for PrivacyPass.
```
exarkun, line 157: It looks like `make_passes` still believes it is receiving the message as `unicode` (it tries to encode the message using UTF-8).

src/_zkapauthorizer/tests/privacypass.py
```diff
@@ -64,10 +64,12 @@ def make_passes(signing_key, for_message, random_tokens):
         for verification_key in verification_keys
     )
     passes = list(
```
exarkun, line 66: `passes` is now a list of `bytes` and gets returned but the docstring still says the return type is `list[unicode]`

src/_zkapauthorizer/model.py
```diff
@@ -1014,7 +1034,7 @@ def to_json_v1(self):
 @attr.s(frozen=True)
 class Voucher(object):
     """
-    :ivar unicode number: The text string which gives this voucher its
+    :ivar unicode bytes: The text string which gives this voucher its
```
exarkun, line 1037: editing error in the change to bytes here

src/_zkapauthorizer/model.py
```diff
@@ -871,7 +891,7 @@ class RandomToken(object):
 
     token_value = attr.ib(
         validator=attr.validators.and_(
-            attr.validators.instance_of(unicode),
+            attr.validators.instance_of(bytes),
```
exarkun, line 894: The docstring for `token_value` needs an update too

exarkun (approved):
Thanks.  Some comments inline, nothing deep - just cases where I think bytes/unicode conversion isn't quite lining up (but mostly places Python 2 lets through).  Please address those and merge.

For what it's worth, I was hoping this would shake out a little bit less evenly:

```
$ git diff origin/main...tom/bytes  | grep -E '^\-.*\.(decode|encode)\(' | wc -l
28
$ git diff origin/main...tom/bytes  | grep -E '^\+.*\.(decode|encode)\(' | wc -l
27
```

On one hand, representing cryptographic inputs as bytes (rather than unicode) does make more sense to me.  On the other hand, they're still being represented as base64-encoded bytes which I feel like subverts any theoretical basis for this preference...  I don't think we can (or should) do anything about this immediately because of python-challenge-bypass-ristretto's API.  Maybe later.

src/_zkapauthorizer/tests/strategies.py
```diff
@@ -548,19 +548,10 @@ def share_parameters():
...
-            min_size=32,
-            max_size=32,
-        )
-        .map(
-            urlsafe_b64encode,
-        )
-        .map(
-            lambda voucher: voucher.decode("ascii"),
-        )
+    return binary(min_size=32, max_size=32,).map(
```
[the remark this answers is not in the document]
tomprince: Trailing commas are one place where black is not entirely opinionated (see [here](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#the-magic-trailing-comma)).

src/_zkapauthorizer/tests/test_client_resource.py
tomprince: This should be *decode*, since json (as a format, not necessarily the implementation) doesn't support byte strings.

src/_zkapauthorizer/tests/storage_common.py
tomprince: Hmm, not sure why I changed this; reverted.

src/_zkapauthorizer/tests/privacypass.py
tomprince: It is actually a list of `Pass`es. I've updated the docstring to match.

tomprince:
> For what it's worth, I was hoping this would shake out a little bit less evenly:
> 
> ```
> $ git diff origin/main...tom/bytes  | grep -E '^\-.*\.(decode|encode)\(' | wc -l
> 28
> $ git diff origin/main...tom/bytes  | grep -E '^\+.*\.(decode|encode)\(' | wc -l
> 27
> ```

Same. Though, if you exclude `model.py` from those, I get 28 before and 8 after. And all but 2 of those[1] after are for vouchers, which is why I was less sure of that part of the change. So, the net effect is to limit the number of places that need to deal with encoding, even if there are still the same number of lines that do. It also appears that 7 places outside of `model.py` that aren't just rewrapped code are all dealing with parsing or generating json, so dealing with encoding there seems reasonable (where as none of the places outside of `model.py` before are clearly at system boundaries).

[1] And one of the remaining 2 was pre-existing but just got rewrapped by black due to other changes in the expression.

src/_zkapauthorizer/tests/privacypass.py
tomprince: Actually, this comment (and thus the change I made to the docstring) is incorrect, and i was confusing what changed here with #269.

PR #265, opened 2021-12-06
Add a metric for invalid ZKAP attempts

### LINKED ISSUE #260: Publish the number of ZKAPs rejected (if we have multiple reasons for rejection, publish a number for each; right now we only have signature errors).

exarkun:
Fixes #260 

This does not expose the total number of ZKAPs that are considered invalid.  Instead it just counts the number of operations attempted where an invalid ZKAP is observed.  Maybe this will provide sufficient visibility.  There is still only one way we can recognize a pass as invalid - invalid signature - so there's only one label on this metric.  It should be straightforward to add more ways as we define them.


exarkun:
Maybe a good idea but ... maybe not?  No demonstrated utility yet, anyway.  Also it ended up being a lot more complex than I anticipated.  Maybe this is something to look into in the future but I'm closing it for now so that the list of open PRs reflects active development efforts.

PR #269, opened 2021-12-14
Replace the direct usages of `make_passes` with `get_passes`.

tomprince:
The later function takes a count of passes to generated, and passes
them to `make_passes`. This also changes `make_passes` to return `Pass`es,
which is what `get_passes` wants to receive.

As part of working on https://whetstone.privatestorage.io/privatestorage/PrivateStorageio/-/issues/84 I want access to the tokens that are generated by `make_passes`/`get_passes`; by using a function that returns `Pass`, I can avoid needing to parse the result to make assertions.


exarkun (approved):
Looks good, thanks.

PR #272, opened 2021-12-17
Defend against the length-contraction slot_testv_and_readv_and_writev DoS

### LINKED ISSUE #222: `slot_testv_and_readv_and_writev` can cause denial-of-service without payment.
We charge users for mutable files based on the logical size of sharefiles. However, the files we actually use on disk can never shrink. In addition, users can cause those files to grow arbitrarily large without ever paying the cost for that size, by truncating the sharefile in the same call; the user is only charged for the truncated size.

exarkun:
Fixes #222

### DIFF
--- a/src/_zkapauthorizer/tests/strategies.py
+++ b/src/_zkapauthorizer/tests/strategies.py
@@ -917,13 +917,13 @@ def slot_test_and_write_vectors():
         TestAndWriteVectors,
         slot_test_vectors(),
         slot_data_vectors(),
-        one_of(
-            # The new length might be omitted completely.
-            just(None),
-            # Or it might be given as an integer.  Allow a zero size which
-            # means "delete this share" in this context.
-            sizes(),
-        ),
+        # The underlying Tahoe-LAFS storage protocol allows None or an integer
+        # here (new_length) to set the size of the mutable container.  The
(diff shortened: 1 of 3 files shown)


hacklschorsch:
LGTM!

src/_zkapauthorizer/tests/strategies.py
```diff
@@ -917,13 +917,13 @@ def slot_test_and_write_vectors():
...
-            # means "delete this share" in this context.
-            sizes(),
-        ),
+        # The underlying Tahoe-LAFS storage protocol allows None or an integer
+        # here (new_length) to set the size of the mutable container.  The
+        # real Tahoe-LAFS storage client never uses this feature and always
+        # passes None.  There are undesirable interactions between non-None
+        # values for new_length and our spending protocol.  Therefore, we
+        # disable non-None values.  So don't generate any here.
+        just(None),
```
tomprince, line 926: Given that this value is never provided by the official client, would it make sense to propose removing support for non-`None` values upstream (similarly to how non-`eq` test operations were removed)?

tomprince (approved):

src/_zkapauthorizer/tests/strategies.py
exarkun: worth pursuing, yea.

PR #273, opened 2021-12-17
Fix certificate validation errors during environment setup on CI

exarkun:
See, for example, https://app.circleci.com/pipelines/github/PrivateStorageio/ZKAPAuthorizer/890/workflows/8e3c1f4b-2c2c-446c-8a37-9e77c5cdd9f6/jobs/2408

```
cachix: ConnectionError (HttpExceptionRequest Request {
  host                 = "cachix.org"
  port                 = 443
  secure               = True
  requestHeaders       = [("Accept","application/json;charset=utf-8,application/json"),("Authorization","<REDACTED>"),("User-Agent","cachix 0.6.0")]
  path                 = "/api/v1/cache/privatestorage-opensource"
  queryString          = ""
  method               = "GET"
  proxy                = Nothing
  rawBody              = False
  redirectCount        = 10
  responseTimeout      = ResponseTimeoutDefault
  requestVersion       = HTTP/1.1
}
 (InternalException (HandshakeFailed (Error_Protocol ("certificate has unknown CA",True,UnknownCa)))))

Exited with code exit status 1

CircleCI received exit code 1
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -148,7 +148,7 @@ jobs:
       # one day someone pushed a bad revision to it and our CI broke.  So now
       # we just pin some recent version.  Who would have thought a floating
       # dependency would cause build instability?
-      - image: "nixos/nix:2.3"
+      - image: "nixos/nix:2.3.16"
 
     # Tahoe-LAFS requires more memory than we get from the default resource
     # class and sometimes we have to build it.
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -164,14 +164,13 @@ jobs:
       # This pin has no particular bearing on what version of our dependencies
       # we are testing against, what version of Python we support, etc.  It is
       # part of CI infrastructure.
-      NIX_PATH: "nixpkgs=https://github.com/NixOS/nixpkgs/archive/28abc4e43a24d28729509e2d83f5c4f3b3418189.tar.gz"
+      NIXPKGS: "https://github.com/NixOS/nixpkgs/archive/28abc4e43a24d28729509e2d83f5c4f3b3418189.tar.gz"
 
     steps:
       - run:
           name: "Set up Cachix"
-          shell: "nix-shell -p cachix bash"
           command: |
-            set -eo pipefail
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -144,8 +144,11 @@ jobs:
         type: "string"
 
     docker:
-      # Run in a highly Nix-capable environment.
-      - image: "nixos/nix:latest"
+      # Run in a highly Nix-capable environment.  We used to use `latest` but
+      # one day someone pushed a bad revision to it and our CI broke.  So now
+      # we just pin some recent version.  Who would have thought a floating
+      # dependency would cause build instability?
+      - image: "nixos/nix:2.3"
```
tomprince, line 151: This doesn't seem like a particularly recent version. Looking at https://hub.docker.com/r/nixos/nix/tags suggests 2.3.12 or 2.3.16 as recent version. It looks like the new images are built with nix, but the older ones are based on https://github.com/NixOS/docker which appears to be apline with nix installed. I think both the versions I mention are that old style.

exarkun:
Empiricism says this is good now...  I'm merging.

.circleci/config.yml
```diff
@@ -170,7 +170,7 @@ jobs:
       - run:
           name: "Set up Cachix"
           command: |
-            nix-env -f $NIXPKGS -iA nixpkgs.cachix nixpkgs.bash
+            nix-env -f $NIXPKGS -iA cachix bash
```
tomprince, line 173: To answer the question in the commit message, `nix-env` by default looks at `~/.nix-defexpr` to get the expression to use (with `-A`, the arguments refer to attributes of this expression). `.nix` files get turned into attributes with that name, and subdirectories with `default.nix` get turned into attributes based on the directory name. Typically, there will be a `nixpkgs` directory (which does have a `default.nix`), which is where the `nixpkgs.<attr>` comes from.

However, `-f <file or url>` causes `nix-env` to use the specified path to be used as an expression (for a directory or a URL (which must have a tarball), it uses the default.nix in it as the expression). Since we are pointing directly at nixpkgs when we pass `-f`, we don't need (and also can't specify) the `nixpkgs.` prefix.

PR #277, opened 2022-01-03
Refactor the Nix packaging for reduced duplication and to provide a more useful shell environment

exarkun:
This changes `default.nix`, `tests.nix`, and `shell.nix` so that we don't need to duplicate `default.nix`'s parameters anywhere yet we can still pass them through from everywhere.

It also adds the lint tools to the environment defined by `shell.nix` for easier interactive linting during development.

It increases the version of Nix required to 2.5 (maybe 2.4 works but I don't see an easy way to get 2.4 to test it out so let's just say 2.5) for `shell.nix` and `tests.nix` but not for `default.nix`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -152,7 +152,7 @@ jobs:
       # This pre-release image in particular fixes a problem in the 2.5.1
       # image where no CA certificates are available.
       # https://github.com/NixOS/nix/issues/5797
-      - image: "nixos/nix:2.6.0pre20211228_f7d22f4-"
+      - image: "nixos/nix:2.5.1"
 
     # Tahoe-LAFS requires more memory than we get from the default resource
     # class and sometimes we have to build it.
@@ -165,6 +165,7 @@ jobs:
       # Let us use features marked "experimental".  For example, most/all of
       # the `nix <subcommand>` forms.
       NIX_CONFIG: "experimental-features = nix-command"
+      NIX_SSL_CERT_FILE: "/nix/var/nix/profiles/default/etc/ssl/certs/ca-bundle.crt"
 
       # Pin a NixOS 21.11 revision.  Most of the software involved in the
       # build process is pinned by nix/sources.json with niv but a few things
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -148,7 +148,7 @@ jobs:
       # one day someone pushed a bad revision to it and our CI broke.  So now
       # we just pin some recent version.  Who would have thought a floating
       # dependency would cause build instability?
-      - image: "nixos/nix:2.3.16"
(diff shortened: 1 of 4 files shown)


shell.nix
```diff
@@ -1,17 +1,14 @@
+# Note: Passing arguments through here to customize the environment does not
+# work on Nix 2.3.  It works with Nix 2.5.  I'm not sure about 2.4.
```
tomprince, line 2: I guess this and `tests.nix` aren't consumed by `PrivateStorageio` so requiring nix 2.4+ isn't an issue for them.

default.nix
```diff
@@ -6,6 +6,7 @@ in
 , mach-nix ? import sources.mach-nix { inherit pkgs pypiData; }
 , tahoe-lafs-source ? "tahoe-lafs"
 , tahoe-lafs-repo ? sources.${tahoe-lafs-source}
+, ...
```
tomprince, line 9: Does this still work with nix 2.3  (given that both nixos 21.05 and 21.11 use that version by default, we either need to support it for this file, or update what we deploy with in PrivateStorageio). I'm not opposed to bumping the version of nix used in our deployment, but if we do, we should probably verify that before making a change here that depends on it.

tomprince (approved):
This looks fine, assuming `default.nix` still work with nix 2.3.

.circleci/config.yml
```diff
@@ -158,6 +162,10 @@ jobs:
       # CACHIX_AUTH_TOKEN is manually set in the CircleCI web UI and allows us to push to CACHIX_NAME.
       CACHIX_NAME: "privatestorage-opensource"
 
+      # Let us use features marked "experimental".  For example, most/all of
+      # the `nix <subcommand>` forms.
+      NIX_CONFIG: "experimental-features = nix-command"
```
tomprince, line 167: Ah, I guess we use `nix path-info` below, which is why we need this. We could consider using `nix build -v -f path/to/file.nix` for the build as well.

default.nix
exarkun: Still just trying to figure out if it is possible to get a working configuration on CI without rolling our own Docker images or something nasty like that. :/

exarkun: For what it's worth, the thing that is broken on Nix 2.3 is that you cannot pass an unnamed argument through `...` from the command line.  So, for example, on Nix 2.3 you cannot override `tahoe-lafs-source` in `default.nix` when running `tests.nix`.

I also had a `python` argument in `default.nix` at one point (not in this branch, probably in the `python3` branch) which would be ignored.  I don't know if we're going to end up supporting more than one version of Python any time soon (since we'll drop Python 2 in the `python3` branch) so this might end up being not a particularly important use-case.

PR #282, opened 2022-01-10
Depend on Tahoe-LAFS 1.17.1

### LINKED ISSUE #280: Tahoe-LAFS 1.17.1 StorageServer is not compatible with our usage
In Tahoe-LAFS 1.17.1 `allmydata.storage.server.StorageServer` no longer implements `RIStorageServer`.  Instead, it implements an unnamed interface which can be used as the protocol-agnostic backend for an `RIStorageServer` implementation.

The `RIStorageServer` implementation is now at `allmydata.storage.server.FoolscapStorageServer`.

ZKAPAuthorizer needs to be adjusted to account for this change.

exarkun:
Stacked on #278 
Fixes #280

### DIFF
--- a/src/_zkapauthorizer/_plugin.py
+++ b/src/_zkapauthorizer/_plugin.py
@@ -140,7 +140,8 @@ class ZKAPAuthorizer(object):
             registry=registry,
         )
         storage_server = ZKAPAuthorizerStorageServer(
-            anonymous_storage_server,
+            # unwrap the Foolscap layer, we'll do it ourselves.
+            anonymous_storage_server._server,
             pass_value=pass_value,
             signing_key=signing_key,
             spender=spender,
--- a/src/_zkapauthorizer/_plugin.py
+++ b/src/_zkapauthorizer/_plugin.py
@@ -140,8 +140,7 @@ class ZKAPAuthorizer(object):
             registry=registry,
         )
         storage_server = ZKAPAuthorizerStorageServer(
-            # unwrap the Foolscap layer, we'll do it ourselves.
-            anonymous_storage_server._server,
+            anonymous_storage_server,
             pass_value=pass_value,
             signing_key=signing_key,
             spender=spender,
--- a/src/_zkapauthorizer/_storage_server.py
+++ b/src/_zkapauthorizer/_storage_server.py
@@ -32,7 +32,7 @@ from typing import Dict, List, Optional
 import attr
 from allmydata.interfaces import RIStorageServer, TestAndWriteVectorsForShares
 from allmydata.storage.common import storage_index_to_dir
-from allmydata.storage.immutable import ShareFile
+from allmydata.storage.immutable import ShareFile, FoolscapBucketWriter
 from allmydata.storage.lease import LeaseInfo
 from allmydata.storage.mutable import MutableShareFile
 from allmydata.storage.server import StorageServer
@@ -253,7 +253,7 @@ class ZKAPAuthorizerStorageServer(Referenceable):
         Pass-through without pass check to allow clients to learn about our
         version and configuration in case it helps them decide how to behave.
         """
-        return self._original.remote_get_version()
+        return self._original._server.get_version()
 
     def remote_allocate_buckets(
         self,
@@ -302,7 +302,7 @@ class ZKAPAuthorizerStorageServer(Referenceable):
             allocated_size,
         )
 
-        alreadygot, bucketwriters = self._original._allocate_buckets(
+        alreadygot, bucketwriters = self._original._server.allocate_buckets(
             storage_index,
             renew_secret,
             cancel_secret,
@@ -338,7 +338,10 @@ class ZKAPAuthorizerStorageServer(Referenceable):
             self._public_key,
             validation.valid[:spent_passes],
         )
-        return alreadygot, bucketwriters
+        return alreadygot, {
+            k: FoolscapBucketWriter(bw)
+            for (k, bw) in bucketwriters.items()
+        }
 
     def remote_get_buckets(self, storage_index):
         """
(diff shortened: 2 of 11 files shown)


src/_zkapauthorizer/_plugin.py
```diff
@@ -140,7 +140,8 @@ def get_storage_server(
             registry=registry,
         )
         storage_server = ZKAPAuthorizerStorageServer(
-            anonymous_storage_server,
+            # unwrap the Foolscap layer, we'll do it ourselves.
+            anonymous_storage_server._server,
```
tomprince, line 144: Based on running the PrivateStorageio system tests, this is incorrect, as `anonymous_storage_server` is already a `StorageServer`, not a `FoolscapStorageServer`.

tomprince (changes requested):

src/_zkapauthorizer/_plugin.py
exarkun: Thanks for this catch.  I'm surprised this is what Tahoe is doing now ... but since it is easier to deal with than what I expected, I guess I won't complain. :slightly_smiling_face:

src/_zkapauthorizer/_storage_server.py
```diff
@@ -211,6 +226,23 @@ def _get_public_key(self):
         # so that `self._signing_key` will be assigned when this runs.
         return PublicKey.from_signing_key(self._signing_key)
 
+    def _bucket_writer_closed(self, bw: BucketWriter):
```
tomprince, line 229: It may be worth a comment that this is duplicated from `FoolscapStorageServer`

src/_zkapauthorizer/_storage_server.py
```diff
@@ -302,7 +334,7 @@ def remote_allocate_buckets(
             allocated_size,
         )
 
-        alreadygot, bucketwriters = self._original._allocate_buckets(
+        alreadygot, bucketwriters = self._original.allocate_buckets(
```
tomprince, line 337: Yay for not not needing to access private methods anymore.

src/_zkapauthorizer/tests/foolscap.py
```diff
@@ -33,13 +32,13 @@ def echo(argument=Any()):
         return Any()
 
 
-@implementer(RIStorageServer)
-class StubStorageServer(object):
-    pass
+class StubStorageBackend(object):
```
tomprince, line 35: I'd be inclined to keep the name, given that this is a stub `StorageServer` implementation.

src/_zkapauthorizer/tests/fixtures.py
```diff
@@ -28,27 +28,28 @@
...
-    clock = attr.ib()
+    clock: Clock = attr.ib()
+
+    tempdir: FilePath = attr.ib(default=None)
+    backend: StorageServer = attr.ib(default=None)
 
     def _setUp(self):
         self.tempdir = FilePath(self.useFixture(TempDir()).join(u"storage"))
-        self.storage_server = StorageServer(
+        self.backend = StorageServer(
```
tomprince, line 52: Similarly, I'm not sure renaming this makes sense.

src/_zkapauthorizer/tests/test_storage_protocol.py
```diff
@@ -154,13 +154,13 @@ def setUp(self):
         )
 
         self.clock = Clock()
-        self.anonymous_storage_server = self.useFixture(
+        self.storage = self.useFixture(
             AnonymousStorageServer(self.clock),
-        ).storage_server
+        )
```
tomprince, line 159: I don't see any uses of this attribute that don't refer to `.backend`/`.storage_server`, so I'd be inclined to keep the test attribute referring to the `StorageServer` attribute (though, I guess it'd make sense to drop the `anonymous_` prefix here).

src/_zkapauthorizer/tests/test_storage_server.py
```diff
@@ -193,15 +194,15 @@ def setUp(self):
         # the same time so we can do lease expiration calculations more
         # easily.
         self.clock.advance(time())
-        self.anonymous_storage_server = self.useFixture(
+        self.storage = self.useFixture(
             AnonymousStorageServer(self.clock),
-        ).storage_server
+        )
```
tomprince, line 199: Same comments apply here.

src/_zkapauthorizer/tests/test_storage_server.py
```diff
@@ -221,7 +222,11 @@ def setup_example(self):
         # Hypothesis and testtools fixtures don't play nicely together in a
         # way that allows us to just move everything from `setUp` into this
         # method.
-        cleanup_storage_server(self.anonymous_storage_server)
+        cleanup_storage_server(self.storage.backend)
+        # One of the tests makes the server read-only partway through.  Make
+        # sure that doesn't leak into other examples.
+        self.storage.backend.readonly_storage = False
```
tomprince, line 228: This feels like it should be part of `cleanup_storage_server` then?

tomprince (approved):
This looks good. There are a couple of minor things to change in the tests (mostly around naming to align with upstream), but is good to merge after that. (I assume that most of the coverage annotation things are due to codecov not working properly)

src/_zkapauthorizer/tests/test_storage_protocol.py
exarkun: I think the `anonymous` is a good component since this attribute lives alongside `self.server` which is our ZKAPAuthorizerStorageServer instance.  Maybe `anonymous_server`?

I guess I'm okay with the "server" part since Tahoe is using that word ... but it's less of a server in 1.17.1.  It used to be a thing you could speak Foolscap at (I guess .. sort of .. you needed a Broker and stuff ...) whereas now it's just a "regular" Python object with regular methods.

tomprince: Oh, I guess it isn't clear from the diff included here, but it was named `annonymous_storage_server`. I was proposing naming `storage_server`, given that the type of the object it is storing is `StorageServer`.

PR #285, opened 2022-01-11
Also test against Tahoe-LAFS master on CI

### LINKED ISSUE #281: Add CI against Tahoe-LAFS master@HEAD
ZKAPAuthorizer is tightly integrated with Tahoe-LAFS.  More than that, Tahoe-LAFS promises no stable Python API but ZKAPAuthorizer consumes one anyway.  This makes ZKAPAuthorizer above-averagely sensitive to Tahoe-LAFS changes.

Rather than waiting for a Tahoe-LAFS release and then reacting to it, we can be more proactive and monitor the state of compatibility master@HEAD revisions of Tahoe-LAFS between releases.  This gives us a better chance of being ready when a release does happen and gives us a chance guide changes in Tahoe-LAFS along a path that's more favorable for ZKAPAuthorizer compatibility *before* those changes are included in a published release.

exarkun:
Fixes #281

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -49,6 +49,10 @@ To update to the current master@HEAD revision, run:
 
    nix-shell --run 'niv update tahoe-lafs-master --branch master'
 
+We intend for these updates to be performed periodically.
+At the moment, they must be performed manually.
+It might be worthwhile to `automate this process <https://github.com/PrivateStorageio/ZKAPAuthorizer/issues/287>` in the future.
+
 It is also possible to pass ``pull/<pr-number>/head`` to test against a specific PR.
 
 The Nix package depends on a version of Tahoe-LAFS determined by the "tahoe-lafs" niv entry.
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -50,12 +50,17 @@ To update to the current master@HEAD revision, run:
    nix-shell --run 'niv update tahoe-lafs-master --branch master'
 
 It is also possible to pass ``pull/<pr-number>/head`` to test against a specific PR.
-When feasible we also test against a released version of Tahoe-LAFS.
+
+The Nix package depends on a version of Tahoe-LAFS determined by the "tahoe-lafs" niv entry.
+When feasible this is a released version of Tahoe-LAFS.
 To update to a new release, run:
 
 .. code:: shell
 
    nix-shell --run 'niv update --rev tahoe-lafs-A.B.C tahoe-lafs'
+
+When it is not feasible to use a released version of Tahoe-LAFS,
+niv's ``--branch`` or ``--rev`` features can be used to update this dependency.
 
 .. note::
 
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -42,6 +42,26 @@ To update the PyPI snapshot (and thus our python dependencies), run
 tahoe-lafs
 ..........
 
+ZKAPAuthorizer declares a dependency on Tahoe-LAFS with a narrow version range.
+This means that Tahoe-LAFS will be installed when ZKAPAuthorizer is installed.
+It also means that ZKAPAuthorizer exerts a great deal of control over the version of Tahoe-LAFS chosen.
+
+When installing using native Python packaging mechanisms
+(for example, pip)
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.rst
```diff
@@ -42,22 +42,28 @@ To update the pypy snapshot (and thus our python dependencies), run
...
-   nix-shell --run 'niv update tahoe-lafs --branch master'
+   nix-shell --run 'niv update tahoe-lafs-master --branch master'
 
 It is also possible to pass ``pull/<pr-number>/head`` to test against a specific PR.
+When feasible we also test against a released version of Tahoe-LAFS.
+To update to a new release, run:
+
+.. code:: shell
+
+   nix-shell --run 'niv update --rev tahoe-lafs-A.B.C tahoe-lafs'
```
tomprince, line 58: This isn't entirely accurate, as this isn't *just* a version we test against, but also the version we run on storage nodes.

tomprince (changes requested):
I think for this to be useful, we need to either update the pin when running CI, or at least have a periodic job that does that (and possibly pushes the result if CI passes). Otherwise, I'm not sure what having a separate niv entry does, that we couldn't get by updating the current one and pushing to CI.

I think I preference would be the third option (run a job in CI periodically, that update the pin and pushed that change if CI passes).

exarkun:
> I think for this to be useful, we need to either update the pin when running CI, or at least have a periodic job that does that (and possibly pushes the result if CI passes). Otherwise, I'm not sure what having a separate niv entry does, that we couldn't get by updating the current one and pushing to CI.

My plan for now is to manually do the update once a week (similar to how we do the nixpkgs update elsewhere).  I'm not sure I want to build GitHub-based automation for this update since ideally we'll move the project to GitLab.

exarkun:
> > I think for this to be useful, we need to either update the pin when running CI, or at least have a periodic job that does that (and possibly pushes the result if CI passes). Otherwise, I'm not sure what having a separate niv entry does, that we couldn't get by updating the current one and pushing to CI.
> 
> My plan for now is to manually do the update once a week (similar to how we do the nixpkgs update elsewhere). I'm not sure I want to build GitHub-based automation for this update since ideally we'll move the project to GitLab.

To add to this, I

exarkun:
I pushed some doc additions to record the current and future plans.

CONTRIBUTING.rst
```diff
@@ -49,14 +49,23 @@ To update to the current master@HEAD revision, run:
...
    nix-shell --run 'niv update tahoe-lafs-master --branch master'
 
+We intend for these updates to be performed periodically.
+At the moment, they must be performed manually.
+It might be worthwhile to `automate this process <https://github.com/PrivateStorageio/ZKAPAuthorizer/issues/287>` in the future.
+
 It is also possible to pass ``pull/<pr-number>/head`` to test against a specific PR.
-When feasible we also test against a released version of Tahoe-LAFS.
+
+The Nix package depends on a version of Tahoe-LAFS determined by the "tahoe-lafs" niv entry.
```
tomprince, line 58: I know that "nix packaging" == "what is deployed on storage nodes", but it might make sense to be explicit about this.

Maybe something like "The package deployed on our storage node depends on the version of tahoe indicated by the `tahoe-lafs` niv entry..."?

tomprince: Maybe it would also make sense to have this section first, given that it is the one that is consumed outside of CI.

CONTRIBUTING.rst
```diff
@@ -42,22 +42,37 @@ To update the pypy snapshot (and thus our python dependencies), run
...
+niv's ``--branch`` or ``--rev`` features can be used to update this dependency.
+
 .. note::
 
    Since tahoe-lafs doesn't have correct version information when installed from a github archive,
    the packaging in ``default.nix`` includes a fake version number.
    This will need to be update manually at least when the minor version of tahoe-lafs changes.
 
-If you want to test multiple versions, you can add an additional source, pointing at other version
+If you want to test additional versions, you can add an additional source, pointing at other version.
```
tomprince, line 75: I think I'd be inclined to put the `tahoe-lafs-master` instructions after this, partly as an example of testing against additional versions.

tomprince (approved):

PR #290, opened 2022-02-01
`niv update tahoe-lafs-master`

exarkun:
Weekly Tahoe-LAFS master CI pin update

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -72,7 +72,8 @@ in
           pyrsistent
           configparser
           eliot
-          foolscap >= 21.7.0
+          foolscap
+          collections-extended >= 2.0.2
 
           # undetected cryptography build dependency
           # https://github.com/DavHau/mach-nix/issues/305
--- a/nix/sources.json
+++ b/nix/sources.json
@@ -71,10 +71,10 @@
         "homepage": "https://tahoe-lafs.org/",
         "owner": "tahoe-lafs",
         "repo": "tahoe-lafs",
-        "rev": "766a53b5208c03c45ca0a98e97eee76870276aa1",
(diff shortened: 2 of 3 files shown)


default.nix
```diff
@@ -72,7 +72,8 @@ in
           pyrsistent
           configparser
           eliot
-          foolscap >= 21.7.0
+          foolscap
+          collections-extended >= 2.0.2
```
exarkun, line 76: Tahoe-LAFS master@HEAD now has an environment marker in the collections-extended requirement declaration.

nix/sources.json
```diff
@@ -29,10 +29,10 @@
         "homepage": "",
         "owner": "DavHau",
         "repo": "pypi-deps-db",
-        "rev": "856d67ab093a68425c3896ec2961cea3b95ae93f",
-        "sha256": "0a7avm4xvm454gxy4dq0fc19j3f9ik8gf3kjpsqhszfkamrn9y0p",
+        "rev": "a44853469d0d6bc65cd382f24e4e60ff59bd60b2",
```
exarkun, line 32: The required version of collections-extended is very new and not present in the pypi-deps-db version we were using before.

tests.nix
```diff
@@ -36,10 +36,12 @@ let
...
       python = "python39";
+      # Pin these narrowly so that lint rules only change when we specifically
+      # want them to.
       requirements = ''
-        isort
-        black
-        flake8
+        isort == 5.10.1
+        black == 21.12b0
+        flake8 == 4.0.1
```
exarkun, line 44: The version of black in the new pypi-deps-db we're updating to has different requirements which are not automatically satisfied by mach-nix.  Rather than try to fix that problem I looked up the versions of these linters we had before and pinned them.

exarkun:
Codecov report is spurious, maybe Hypothesis got unlucky or something.

PR #293, opened 2022-02-02
Design document for replication and recovery system

exarkun:
This designs but does not implement #235.


docs/source/designs/backup-recovery.rst
```diff
@@ -0,0 +1,515 @@
...
+
+The replication process is as follows:
+
+#. Replication is configured using the external interface.
+
+   #. The *replica directory*,
+      a new mutable directory,
+      is created on grid.
+   #. The write capability is added to the database.
+   #. The read capability is returned to the external caller.
```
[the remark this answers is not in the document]
crwood: Also, if the writecap is contained in the database and that database is later restored, does that mean that the newly-restored instance will (potentially) be writing to the same (replica?) directory as the old device/instance? Can/should additional care be taken to prevent old possessors of the writecap from messing with new/post-restore replicas/snapshots?

docs/source/designs/backup-recovery.rst
```diff
@@ -0,0 +1,515 @@
...
+       If the snapshot does not exist then its size is treated as 0.
+
+.. [2] A snapshot is obsolete if there is a completely uploaded snapshot with a greater sequence number.
+
+.. [3] Application-code is supplied with a cursor which performs this capturing.
+       Replication code bypasses this capturing so that statements which record the event stream are not themselves recorded.
+       Recovery code bypasses this capturing so that statements to recreate the database are also not recorded.
+       ``SELECT`` statements are ignored since they cannot change the database (XXX is this true?).
+
+.. [4] The event stream in the database is large enough when it is larger than 900,000 bytes.
```
crwood, line 484: Is this including or excluding overhead from erasure coding? Can you specify?

docs/source/designs/backup-recovery.rst
```diff
@@ -0,0 +1,515 @@
...
+The recovery process is as follows:
+
+#. An empty database is created.
+#. The snapshot is downloaded.
+#. The event stream is downloaded.
+#. The statements from the snapshot are executed against the database.
+#. The statements from the event stream,
+   starting at the first statement after the snapshot's sequence number,
+   are executed against the database.
+
```
crwood, line 295: To follow a previous comment above, I wonder if caps should also be rotated here to minimize the odds of multiple actors writing to the same mutable/directory.

docs/source/designs/backup-recovery.rst
```diff
@@ -0,0 +1,515 @@
...
+The maintenace of a replica is an ongoing process.
+The replica loses value,
+up to and including *all* value,
+if that maintenance process breaks down at some point.
+
+Ideally it would be possible for some component to detect problems with this process.
+Where possible,
+problems should be corrected automatically.
+At some point the system may determine no automatic correction is possible and user intervention is required.
```
crwood, line 386: I realize that the user interaction part is out of scope here, but how/where will such errors be reported? Perhaps in the body of `GET /storage-plugins/privatestorageio-zkapauthz-v1/replicate`?

docs/source/designs/backup-recovery.rst
```diff
@@ -0,0 +1,515 @@
...
+
+Threat Model
+~~~~~~~~~~~~
+
+This design aims to defend accounting data in the same way user data is defended.
+If the capability for the replica directory is kept confidential then the accounting data will be kept confidential.
+It is up to the party using the external interface to keep the capability confidential.
+
+This system creates new copies of accounting data on the Tahoe-LAFS grid.
+The convenient-related requirements for the user stories at the top of this design imply that the capabilities for accessing user data will grant access to read the accounting data replicas created by this system.
```
crwood, line 422: "convenient-related" -> "convenience-related"

src/_zkapauthorizer/backup-recovery.yaml
```diff
@@ -0,0 +1,136 @@
...
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ReplicaConfiguration"
+
+    post:
+      description: >-
+        Configure an on-grid replica.
+
+        Once a replica has been configured this node will keep that replica
+        up-to-date with local state changes.
```
crwood, line 80: It might be useful for callers to be able to manually/force a replica to be created/uploaded on demand (rather than needing to wait until that object is of "large enough" size -- for example, in cases where a user has just finished uploading all of their files and wants to be sure that their remote database is fully up-to-date before they turn off their laptop.

crwood (commented):
Thanks for writing all of this out! It's nice to see a viable a solution brewing that doesn't require shipping litestream (or depending on tahoe's SFTP server). ;)

I left a few comments/questions throughout below. Do with them what you will.

docs/source/designs/backup-recovery.rst
```diff
@@ -0,0 +1,515 @@
...
+Replication
+~~~~~~~~~~~
+
+The replication process is as follows:
+
+#. Replication is configured using the external interface.
+
+   #. The *replica directory*,
+      a new mutable directory,
+      is created on grid.
```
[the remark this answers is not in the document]
meejah: I think it's best if ZKAPAuthorizer controls the writecap; it's in a much better position to ensure it's the only client writing to it IMO.

docs/source/designs/backup-recovery.rst
meejah: I'm also curious why 900KB was chosen.

docs/source/designs/backup-recovery.rst
meejah: I assumed (but maybe this should be explicit) that a post-recovery instance would need to be told to start doing backups again..? (and then create a new writecap etc)

src/_zkapauthorizer/backup-recovery.yaml
meejah: Perhaps an "even better" use-case would be a user preparing to go offline for a while (or cross a border) wanting to ensure stuff is up-to-date.

Or, a user may be _planning_ to recover later (e.g. they're going to wipe their device, maybe they're crossing a border or giving it back to an employer or to a repair shop).

docs/source/designs/backup-recovery.rst
```diff
@@ -0,0 +1,515 @@
+.. Heading order = - ~ `
+
+ZKAP Database Replication / Recovery
```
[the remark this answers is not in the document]
exarkun: That's an interesting point.  Considering your suggested connotations, I end up feeling like this system actually does both - it recovers state from the grid (it has to find it using the replica readcap and then download it) and then it uses that state to restores the SQLite3 database it needs in order to operate normally (because the on-grid state is not the SQLite3 database itself, just information which can be used to recreate -- or "reconstruct" it, even).

Separate from that, I think there's a different sense of "recovery" which is slightly more common in software systems - the sense of "failure recovery" or "disaster recovery" where the understanding is that this is a thing you do in order to transition a system in a failed state back to a working state.  I'm not sure how widely this sense would be understood outside of software systems ... but maybe that doesn't matter since this is a software system design. :)

However, these conclusions arise from a detailed knowledge of the implementation strategy chosen.  Given that, maybe the name of the overall system being designed doesn't need to reflect them.  It should reflect the value delivered to the consumer of the system.  I think that semantic distinctions between these words is blurry enough that I have a hard time seeing which is a better fit from that external perspective (perhaps partially because I've been inside it too long).

Since it sounds like you don't think this is critical to understanding of the system I'll probably leave it as-is.

docs/source/designs/backup-recovery.rst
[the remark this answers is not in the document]
exarkun: > I'm not sure what is achieved by diminishing this cap. If a) the writecap is in the (remotely-stored?) database and b) the readcap points to that database, then the possessor of the readcap can trivially discover the writecap.

Hmm.  Indeed.  I thought to diminish it just because no one outside of the system ever needs the write cap but it is true that this readcap can be leveraged to obtain the writecap...

I'm not sure what to do about that.  Maybe it is still a good idea to diminish it and then try to figure out a way to remove the property that allows an attacker to leverage it into a writecap (maybe later if we can't figure out how to do it now)?  Put another way, "there is a hole over there so we might as well make a same-shaped hole over here" doesn't make sense, even though having one hole is still bad.

> Also, if the writecap is contained in the database and that database is later restored, does that mean that the newly-restored instance will (potentially) be writing to the same (replica?) directory as the old device/instance?

Hm.  I considered this a feature of the design, not a flaw.  But maybe that was a mistake.

If we want the newly-restored instance to use a different replica we can arrange that, I think.  For example, we could say that a newly-restored instance forgets its replication configuration and must be re-configured.  Generate a new replica dircap, re-upload (maybe if we're really careful the "re-upload" just gives you the same immutable caps but in a new location, or maybe we can take explicit steps to make sure that is what happens ... but that's just an optimization), proceed to replicate to the new location, the old one is forgotten.

Maybe this even suggests that the replica writecap is either not kept in the database or is not replicated along with the rest of the database state - which closes the above-mentioned hole, too.

docs/source/designs/backup-recovery.rst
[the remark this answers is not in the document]
exarkun: If we think of a motivation for externally-supplied writecap we can adjust the API or add another one in the future.  I agree with @meejah that a nice property to get out of the system is that in the course of normal operation the writecap isn't shown to anyone/anything outside of ZKAPAuthorizer so it is harder to achieve *casual* breakage through concurrent writes.  Of course someone can dig through internal ZKAPAuthorizer state to find it and then break stuff but at this point it's basically their responsibility.

docs/source/designs/backup-recovery.rst
exarkun: Are there backup/recovery user stories/acceptance criteria already written somewhere that talks about expectations about data read/write access?

docs/source/designs/backup-recovery.rst
exarkun: Absent a user-interaction design that covers exposing this information, it will probably just end up in the Tahoe-LAFS client node log.

src/_zkapauthorizer/backup-recovery.yaml
exarkun: Maybe it should synchronize the replica before allowing the Tahoe-LAFS client node process to exit?  Then every time you cleanly quit the app the replica is up to date.

docs/source/designs/backup-recovery.rst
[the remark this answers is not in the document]
meejah: At least returning a read-only cap to the caller makes it more-obvious that they shouldn't be writing to it .. leaking the read-cap is still bad (and I guess leaking the write-cap is _even worse_ since you could then wipe the data).

> For example, we could say that a newly-restored instance forgets its replication configuration and must be re-configured.

I think this is the right thing, to avoid the situation where the "old device" is found/used again (and then you have multiple writers to the same writecap).

docs/source/designs/backup-recovery.rst
meejah: Ignoring the user entirely: it's much easier to reason about "will we have precisely one writer to this writecap" (i.e. meet the Tahoe constraints) if a new one is created for each "instance" of zkap-authorizer (even if it's a recovery of another instance).

src/_zkapauthorizer/backup-recovery.yaml
meejah: I know this was discussed elsewhere, and I certainly agree that could be a good Tahoe feature .. but I think it's "complex" to get correct there, even with only a single plugin (e.g. zkap-authorizer needs to be shut down _first_ and ~all the other Tahoe services need to be alive while it does so).

So, both pragmatically and under the "explicit is better than implicit" banner, an endpoint to ask zkapauthorizer in particular to do this would be good. (However, under the "it's complex" part, then "something" -- zkap or gridsync -- needs to wait until Tahoe is done).

Perhaps a hybrid solution: Tahoe grows a feature where a "nice" exit signal means it finishes all "active operations" _and_ zkapauthorizer has an explicit endpoint to "backup all outstanding data, now" ... so then gridsync calls the zkap-authorizer endpoint (queuing up any required active Tahoe operations), then kills tahoe, then waits for Tahoe to exit.

As another "plus" for an explicit endpoint: it may be that a user _doesn't_ care about the data being in the grid (e.g. a "normal close of the app") and just wants a quick but orderly shutdown. (I'm not trying to actually design the UX here, but it's a lot more flexible and "up to the UI" with the endpoint versus implicit stuff like "I hope that TERM means my data got there").

exarkun: I added a note about the "safe border crossing" use-case to the roadmap doc.

docs/source/designs/backup-recovery.rst
exarkun: I elaborated on this and fixed the doc to describe how to reach the pre-encoding threshold (which is actually 570,000 bytes for our current grid configuration).

PR #295, opened 2022-02-07
Implement a limited form of the "recovery" endpoint of the replication/recovery design

exarkun:
This implements recovery *only* from a snapshot (there is no event-stream handling) from the new replication/recovery design.


hacklschorsch:
Good stuff so far :)

exarkun:
This is being broken into merge-sized pieces.  It might be interesting as a reference or something but it will never be merged itself.

PR #296, opened 2022-02-07
Clean up used-up tokens from database

### LINKED ISSUE #288: The `tokens` table never has tokens deleted from it
This is the table holding "random tokens".  It is populated by `VoucherStore.add`.  Signatures are cleaned up from the corresponding `unblinded-tokens` table
* by `VoucherStore.discard_unblinded_tokens` after they're spent
*  by `VoucherStore.invalidate_unblinded_tokens` after a spending failure

However, nothing ever removes rows from the `tokens` table.  This means there's a storage leak that grows with the number of tokens acquired.  There's also potentially a privacy leak since the random tokens are used during the spending process and therefore they create a link between a client and an operation.

exarkun:
Fixes #288

### DIFF
--- a/src/_zkapauthorizer/model.py
+++ b/src/_zkapauthorizer/model.py
@@ -17,17 +17,14 @@ This module implements models (in the MVC sense) for the client side of
 the storage plugin.
 """
 
-from base64 import b64decode
 from datetime import datetime
 from functools import wraps
 from json import loads
 from sqlite3 import OperationalError
 from sqlite3 import connect as _connect
-from typing import List
 
 import attr
 from aniso8601 import parse_datetime
-from challenge_bypass_ristretto import UnblindedToken as _UnderlyingUnblindedToken
 from twisted.logger import Logger
 from twisted.python.filepath import FilePath
 from zope.interface import Interface, implementer
--- a/src/_zkapauthorizer/model.py
+++ b/src/_zkapauthorizer/model.py
@@ -433,6 +433,14 @@ class VoucherStore(object):
                 "Cannot insert tokens for unknown voucher; add voucher first"
             )
 
+        cursor.execute(
+            """
+            SELECT [counter] FROM [vouchers] WHERE [number] = ?
+            """,
+            (voucher.decode("ascii"),),
+        )
+        (new_counter,) = cursor.fetchone()
+
         cursor.executemany(
             """
             INSERT INTO [unblinded-tokens] ([token], [redemption-group]) VALUES (?, ?)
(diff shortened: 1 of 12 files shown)


src/_zkapauthorizer/model.py
```diff
@@ -439,6 +442,64 @@ def insert_unblinded_tokens_for_voucher(
...
+    def _delete_corresponding_tokens(self, cursor, voucher : bytes, unblinded_tokens : List["UnblindedToken"]) -> None:
+        """
+        Delete rows from the [tokens] table corresponding to the given unblinded
+        tokens.
+        """
+        # The only way to match tokens with unblinded tokens is to compare the
+        # preimages they each contain.  Unfortunately this means we have to
+        # load all of the tokens from the database.  Hopefully this will never
+        # be a truly huge number because we clean up the table as we make
+        # progress on redemption.
```
tomprince, line 456: While this is true, looking at the code for `BatchDLEQProof..invalid_or_unblind`, it looks like the code always returns the list of unblinded tokens in the same order as the tokens that were passed in[1]. Thus, it would be possible for the caller to pass in the corresponding list of tokens, rather than requiring a bunch of logic here to determine the list of tokens to delete.

Specifically, if `random_tokens` were passed to `RedemptionResult` [here](https://github.com/PrivateStorageio/ZKAPAuthorizer/blob/ccc0d28fb4f54c10722ee311945073296e557fec/src/_zkapauthorizer/controller.py#L575-L578), they could be passed to `insert_unblinded_tokens_for_voucher` [here](https://github.com/PrivateStorageio/ZKAPAuthorizer/blob/ccc0d28fb4f54c10722ee311945073296e557fec/src/_zkapauthorizer/controller.py#L918-L924), which could then be passed into this function.

[1] I think it is reasonable to depend on this, since that function also depends on the list of blinded and signed tokens to be in the same order. And, in any case, it doesn't actually appear that we would be depending on the order.

---

Alternatively, it looks like the tokens to delete are all the tokens in the database associated with the given vouncher and counter, so, if the counter was passed into `insert_unblinded_tokens_for_voucher` [here](https://github.com/PrivateStorageio/ZKAPAuthorizer/blob/ccc0d28fb4f54c10722ee311945073296e557fec/src/_zkapauthorizer/controller.py#L918-L924), we could just delete tokens based on that info. Looking at 
['PaymentController._get_random_tokens_for_voucher`](https://github.com/PrivateStorageio/ZKAPAuthorizer/blob/ccc0d28fb4f54c10722ee311945073296e557fec/src/_zkapauthorizer/controller.py#L767-L801)
and [`VoucherStore.add`](https://github.com/PrivateStorageio/ZKAPAuthorizer/blob/ccc0d28fb4f54c10722ee311945073296e557fec/src/_zkapauthorizer/model.py#L264-L339) the tokens that will be deleted by the code in this PR will always match the tokens with the given voucher and counter.

tomprince (changes requested):
It looks like we could easily know, at the point that `insert_unblinded_tokens_for_voucher` is called, the list of tokens corresponding to that redemption, so I would be inclined to avoid recalculating them to figure out which ones to delete.

src/_zkapauthorizer/model.py
exarkun: Yea good call.  This was actually plan A (94f8ca027a4a12e1c9543396f90bd274cf7b8c09) but it looks like I picked the wrong column name and then (mistakenly) convinced myself the current schema wouldn't support that approach.

I've switched the branch back to that idea w/ a correct implementation - with one variation on your suggestion.  I just pulled the counter out of the database instead of requiring it as a parameter to `insert_unblinded_tokens_for_voucher`.  The database state has to be consistent with the API calls being made (so the value used should be the same either way) and this way I don't have to update all the callers of `insert_unblinded_tokens_for_voucher`.

src/_zkapauthorizer/model.py
```diff
@@ -422,14 +426,22 @@ def insert_unblinded_tokens_for_voucher(
...
                 "Cannot insert tokens for unknown voucher; add voucher first"
             )
 
+        cursor.execute(
+            """
+            SELECT [counter] FROM [vouchers] WHERE [number] = ?
+            """,
+            (voucher_text,),
+        )
+        (new_counter,) = cursor.fetchone()
```
tomprince, line 443: I guess there isn't a way to get the value set in the update statement two lines up. :upside_down_face:

tomprince (approved):

exarkun:
> I originally thought that if you are doing that, it might make sense to not pass in completed to insert_unblinded_tokens_for_voucher either. Except I guess calculating the number of groups depends on data not available to the store, but could pass in that value instead.

Yep, had similar thoughts.  I think this shows the weakness/incorrectness of some of the layering or abstractions involved here (the "counter" thing was introduced after the initial implementation, maybe not very well).  I vaguely expect that we'll do something to clean up the factoring someday and this signature will get a bit less gross...

PR #299, opened 2022-02-13
Change how success is reported for the replication recovery API

exarkun:
Related to #235.  This changes the how the recovery endpoint reports success.  Instead of synchronously returning when recovery is successful, it will return success when the recovery attempt has begun.  A separate endpoint will exist to allow the state to be polled.


hacklschorsch (approved):
LGTM!

PR #308, opened 2022-02-18
add a basic recovery function and integrate it with StatefulRecoverer

exarkun:
This is related to #235.  It plugs a basic recovery function into `StatefulRecoverer` and reflects its use in the exposed state.


exarkun:
The Windows CI failure is an unrelated intermittent issue with floating point precision. :(

PR #309, opened 2022-02-20
Add a Tahoe-LAFS replica snapshot downloader

exarkun:
This relates to #235.  It integrates the downloader and recoverer together in StatefulRecoverer and implements a Tahoe-LAFS downloader that knows the recovery cap is a directory with a "snapshot.sql" child.

It also adds a bit more to our Tahoe-LAFS HTTP API helper library for creating directories, linking children into them, and retrieving those children again later.

In principle I expect this could actually recover from a snapshot now, except it is not yet plumbed into our HTTP API by `_plugin.py` (and of course nothing uploads a snapshot yet, and there is no "event stream" handling).

### DIFF
--- a/src/_zkapauthorizer/recover.py
+++ b/src/_zkapauthorizer/recover.py
@@ -208,7 +208,10 @@ async def tahoe_lafs_downloader(
     node's private directory.
     """
     api_root = DecodedURL.from_text(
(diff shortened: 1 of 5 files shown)


exarkun:
Filed the Windows failure as #310

Issue #310, opened 2022-02-20
intermittent failure in _zkapauthorizer.tests.test_recover.StatefulRecoverTests.test_recover


exarkun:
```
[FAIL]
Traceback (most recent call last):
Failure: testtools.testresult.real._StringException: Traceback (most recent call last):
  File "C:\hostedtoolcache\windows\Python\3.9.10\x64\lib\site-packages\_zkapauthorizer\tests\test_recover.py", line 150, in test_recover
    run_state_machine_as_test(
  File "C:\hostedtoolcache\windows\Python\3.9.10\x64\lib\site-packages\hypothesis\stateful.py", line 222, in run_state_machine_as_test
    run_state_machine(state_machine_factory)
  File "C:\hostedtoolcache\windows\Python\3.9.10\x64\lib\site-packages\hypothesis\stateful.py", line 107, in run_state_machine
    @given(st.data())
  File "C:\hostedtoolcache\windows\Python\3.9.10\x64\lib\site-packages\hypothesis\core.py", line 1190, in wrapped_test
    raise the_error_hypothesis_found
  File "C:\hostedtoolcache\windows\Python\3.9.10\x64\lib\site-packages\_zkapauthorizer\tests\test_recover.py", line 90, in snapshot_equals_database
    self.case.assertThat(
  File "C:\hostedtoolcache\windows\Python\3.9.10\x64\lib\site-packages\testtools\testcase.py", line 480, in assertThat
    raise mismatch_error
testtools.matchers._impl.MismatchError: !=:
reference = ['BEGIN TRANSACTION;',
 'CREATE TABLE [0] ([0] REAL);',
 'INSERT INTO "0" VALUES(1.12589990684262408748e+15);',
 'CREATE TABLE [1] ([0] INT);',
 'COMMIT;']
actual    = ['BEGIN TRANSACTION;',
 'CREATE TABLE [0] ([0] REAL);',
 'INSERT INTO "0" VALUES(1.12589990684262430953e+15);',
 'CREATE TABLE [1] ([0] INT);',
 'COMMIT;']
: after <function <lambda>> on <sqlite3.Connection object at 0x0000019744DC7E40>: source (reference) database iterdump does not equal sink (actual) database iterdump


_zkapauthorizer.tests.test_recover.StatefulRecoverTests.test_recover
```

[...]

PR #311, opened 2022-02-21
move testresources to a module where we could share them

exarkun:
This is related to #235.  It is just moving code around to more sensible locations.  The branch name is weird because I thought I was gonna write some integration tests on top of this ... but there's too many missing pieces, still, so I'm not gonna right now.


exarkun:
The Windows failure is #310

PR #312, opened 2022-02-21
Compare structured SQL dumps

### LINKED ISSUE #310: intermittent failure in _zkapauthorizer.tests.test_recover.StatefulRecoverTests.test_recover

exarkun:
Eventually this might help fix #310 .  Right now all it does is replace the SQL text comparison in the test suite with comparison of the underlying values, mostly without the SQL (the schema ddl is the exception).

This should enable use of a different floating point comparison logic which could allow the near-epsilon differences that seem to come up to pass through without failing the tests.  That part is not implemented yet.  I cannot reproduce these failures on Linux so I want to see some runs of this code on CI before proceeding.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -257,12 +257,6 @@ workflows:
         # release.
         tahoe-lafs-source: "tahoe-lafs-master"
 
-    # https://circleci.com/docs/2.0/testing-ios/#supported-xcode-versions
-    - "macos-tests":
-        name: "macOS tests python 3.8 xcode 11.7.0"
-        py-version: "3.8"
-        xcode-version: "11.7.0"
-
     - "macos-tests":
         name: "macOS tests python 3.9 xcode 12.3.0"
         py-version: "3.9"
--- a/src/_zkapauthorizer/tests/matchers.py
+++ b/src/_zkapauthorizer/tests/matchers.py
@@ -27,8 +27,10 @@ __all__ = [
 
 from datetime import datetime
 from json import loads
+from math import isnan, nextafter
 
 import attr
+from attrs import define
 from testtools.matchers import (
     AfterPreprocessing,
     AllMatch,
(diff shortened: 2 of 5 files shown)


hacklschorsch (approved):
"LGTM"

Hahah next thing I finally learn SQL proper

exarkun:
Still trying to narrow in on good test coverage...

exarkun:
Superseded by #318

PR #313, opened 2022-02-22
prepare for having a second service

exarkun:
Generalize the existing hack to integrate our lease maintenance service.
This makes it easily usable for more services.


exarkun:
The Windows failure is #310...

PR #315, opened 2022-02-22
add an endpoint for enabling replication

exarkun:
Relates to #235.  This adds the API that a client can use to enable the replication system and learn the recovery capability.  It is not hooked up to anything real yet.


exarkun:
The Windows failure is #310

PR #316, opened 2022-02-22
implement the replication setup function

exarkun:
This relates to #235.  It implements a function which will really configure a ZKAPAuthorizer-enabled Tahoe-LAFS node for replication.  This can be plugged in to the resource introduced by #315 (but it is not yet).


exarkun:
Windows #310 etc

PR #317, opened 2022-02-22
Some Nix-based development environment amenities

exarkun:
This puts the source tree in the environment so `trial _zkapauthorizer` just works.  It also puts some elpy dependencies into the environment so elpy (emacs python support) can find various dev tools and automatically apply them to sources as they're edited.


exarkun:
Windows #310 etc

PR #318, opened 2022-03-04
Fix floating point comparisons in the recovery tests

### LINKED ISSUE #310: intermittent failure in _zkapauthorizer.tests.test_recover.StatefulRecoverTests.test_recover

exarkun:
Fixes #310 

This is like #312 but also adds custom floating point comparison logic which will *allow* the small floating point corruption that SQLite3 introduces in `iterdump` to pass without failing the tests.

Surely this is not as good as avoiding the corruption in the first place but I think that involves designing and using a binary floating point format to replace the `iterdump` format.  We can do that some other time, I guess.

### DIFF
--- a/src/_zkapauthorizer/tests/test_matchers.py
+++ b/src/_zkapauthorizer/tests/test_matchers.py
@@ -18,7 +18,7 @@ Tests for ``_zkapauthorizer.tests.matchers``.
 
 from json import dumps
 from math import isfinite, nextafter
-from sqlite3 import connect
+from sqlite3 import Connection, connect
 
 from hypothesis import assume, example, given
 from hypothesis.strategies import (
@@ -260,26 +260,35 @@ def _float_example(fs):
     )
 
 
+def copy(src_db: Connection) -> Connection:
+    """
+    Return an in-memory SQLite3 database that is a copy of the given database.
+    """
+    db = connect(":memory:")
+    list(map(db.execute, src_db.iterdump()))
+    return db
+
+
 class EqualsDatabase(TestCase):
     """
     Tests for the ``equals_database`` matcher.
     """
 
     def setup_example(self):
-        self.a = connect(":memory:")
-        self.b = connect(":memory:")
+        self.original = connect(":memory:")
 
     @given(sql_schemas())
     def test_same_schema(self, tables):
         """
         Two databases with the same schema match.
         """
-        for db in [self.a, self.b]:
-            for name, table in tables.items():
-                db.execute(create_table(name, table))
+        for name, table in tables.items():
+            self.original.execute(create_table(name, table))
+
+        copied = copy(self.original)
 
         self.assertThat(
-            equals_database(self.a).match(self.b),
+            equals_database(self.original).match(copied),
             Is(None),
         )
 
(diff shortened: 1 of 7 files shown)


src/_zkapauthorizer/tests/test_matchers.py
```diff
@@ -142,3 +164,274 @@ def test_matches(self):
...
+            1.500589370727539,
+            1125899906842624.0,
+            5.192298096474867e33,
+            # This one has a distance of 3
+            -1.1919735848895067e-35,
+        ]
+    )
+    @_float_example(
+        [
+            # Examples from Rick Regan's blog post that are broken on Windows.
```
hacklschorsch, line 400: Maybe add the URL since you don't have his name above where you have the URL:

https://www.exploringbinary.com/incorrect-decimal-to-floating-point-conversion-in-sqlite/

hacklschorsch (approved):
Minor nitpicks at best - and even that only in the comments.
It's a lot of code for an as-of-yet hypothetical scenario, but as always I find your stuff pristine.
Merge at will!

PR #330, opened 2022-03-18
Clean up SQLite3 database connection leaks from the test suite

### LINKED ISSUE #329: The test suite leaks SQLite3 connections
Something in the test suite leaves SQLite3 connections open - I don't know what.  This has probably been the case for a while but now the test suite is large enough that the number of leaked connections gets close to the default Linux open file descriptor limit and sometimes causes failures due to this limit.

exarkun:
Fixes #329

### DIFF
--- a/src/_zkapauthorizer/__init__.py
+++ b/src/_zkapauthorizer/__init__.py
@@ -18,6 +18,12 @@ __all__ = [
     "NAME",
 ]
 
+
+# Hotfix Tahoe-LAFS #3883
+from allmydata import stats
+
+stats.eventually = lambda f: f()
+
 # The identifier for this plugin.  This appears in URLs for resources the
 # client plugin exposes, configuration files, etc.
 NAME = "privatestorageio-zkapauthz-v2"
--- a/default.nix
+++ b/default.nix
@@ -3,6 +3,8 @@ let
 in
 { pkgs ? import sources.release2111 { }
 , pypiData ? sources.pypi-deps-db
+  # You might want more things like tkinter - in which case, you want
+  # pythonXXFull.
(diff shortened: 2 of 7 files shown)


src/_zkapauthorizer/__init__.py
```diff
@@ -18,6 +18,12 @@
     "NAME",
 ]
 
+
+# Hotfix Tahoe-LAFS #3883
+from allmydata import stats
+
+stats.eventually = lambda f: f()
```
exarkun, line 25: This is the actual fix.  The rest is perhaps nice cleanup but ultimately irrelevant to the specific problem described by the related issue.

meejah: Shouldn't it forward args + kwargs? Or does stats just happen to not ever use those...?

(It seems ... surprising .. that this fixes a memory/resource leak)

exarkun: For strict compatibility you're right - but, indeed, allmydata.stats doesn't pass any arguments so it won't make a practical difference (and I put up a Tahoe-LAFS PR to stop using `eventually` here entirely).

Regarding surprisingness ... I can say I was dismayed but not surprised that investigations led here.  Basically, `eventually` is like `reactor.callLater(0, ...)` and `reactor.callLater` takes a reference to its arguments and holds them until they run ... but most of the ZKAPAuthorizer test suite doesn't run the reactor so the references are held for the whole test suite run.  Thus, every `_Client` object the test suite ever creates lives for the full duration of the test suite.  That's not *so* destructive, but the suite is now large enough that the file descriptors for SQLite3 connections that are held open as a result are a problem.

PR #338, opened 2022-03-23
a 'fast' (low-examples) Hypothesis profile

meejah:


exarkun (approved):
As discussed yesterday - indeed, this is a helpful tool for doing quicker runs of the test suite, especially on less than the absolute newest/fastest compute substrate.

Thanks.

Issue #337, opened 2022-03-23
Upload a new snapshot to remote storage 


meejah:
Related to #235 (I can't edit this in, but should be a TODO below the one matching the subject of this ticket).

Given some snapshot data, upload it to the recovery mutable directory.

- create an immutable capability from the data
- put ^ under a static name `snapshot.sql` (overwriting new ones) 
  - if we crash before successfully updating the entry, we'll still have the old one
  - ...and will overwrite it (properly) next time we decide to update

PR #339, opened 2022-03-23
337.upload snapshots

meejah:
#337 (but stacked on #338 too)


exarkun:
It looks like GitHub Actions didn't bother to run ... I suppose something to do with the PR being from a fork.  I don't remember what extra buttons might need to be clicked to fix that. :(

meejah:
will try with a non-fork version

Issue #333, opened 2022-03-21
_zkapauthorizer.recover.recover is incompatible with VoucherStore.call_if_empty


exarkun:
* `recover` requires that a transaction has already been started on the cursor passed in but the snapshot data does its own transaction management (so sqlite3 raises an exception)
* The cursor supplied has foreign key constraints enabled but foreign keys will be invalid at intermediate states during the recovery process (so sqlite3 raises an exception)
* There is certainly already a schema in the database associated with the given cursor but `recover` will try to load the schema from the snapshot (so sqlite3 raises an exception)
* The overall recovery process is asynchronous but `call_if_empty` only works with synchronous functions

PR #340, opened 2022-03-23
Convert `VoucherStore.call_if_empty` to be/support async functions

exarkun:
This is related to #235 and specifically to the last buillet point of #333.  Since recovery is async, `call_if_empty` must support async functions.  This makes `call_if_empty` async and fixes the web resource code to work with this.

It _also_ changes the format of snapshots from bare-SQL strings concatenated to netstring-encoded SQL strings concatenated.  This is necessary to be able to actually recover a snapshot ~which the new unit test that demonstrates `call_if_empty` / `recover` integrate requires~ which a test in a follow-up PR will demonstrate/require.  Otherwise, the dumped SQL statements with embedded newlines get split into pieces and cause syntax errors when recovery tries to execute them.

### DIFF
--- a/src/_zkapauthorizer/model.py
+++ b/src/_zkapauthorizer/model.py
@@ -279,7 +279,7 @@ class VoucherStore(object):
         )
 
     @with_cursor_async
-    def call_if_empty(self, cursor, f: Callable[[Cursor], _T]) -> _T:
+    async def call_if_empty(self, cursor, f: Callable[[Cursor], Awaitable[_T]]) -> _T:
         """
         Transactionally determine that the database is empty and call the given
         function if it is or raise ``NotEmpty`` if it is not.
@@ -293,7 +293,7 @@ class VoucherStore(object):
         # `invalid-unblinded-tokens` table and maybe also look at lease
         # maintenance spending.
         if self.list.wrapped(self, cursor) == []:
-            return f(cursor)
+            return await f(cursor)
         else:
             raise NotEmpty()
 
--- a/src/_zkapauthorizer/model.py
+++ b/src/_zkapauthorizer/model.py
@@ -278,11 +278,14 @@ class VoucherStore(object):
             conn,
         )
 
-    @with_cursor
-    def call_if_empty(self, cursor, f: Callable[[Cursor], None]) -> None:
+    @with_cursor_async
+    def call_if_empty(self, cursor, f: Callable[[Cursor], _T]) -> _T:
         """
         Transactionally determine that the database is empty and call the given
         function if it is or raise ``NotEmpty`` if it is not.
+
+        The function may return an ``Awaitable``.  If it does the transaction
+        opened for it will be kept open until the ``Awaitable`` completes.
         """
         # After redeemed-voucher garbage collection is implemented, this won't
         # be enough of a check.  We should check the unblinded-tokens table
--- a/src/_zkapauthorizer/recover.py
+++ b/src/_zkapauthorizer/recover.py
@@ -128,7 +128,7 @@ class StatefulRecoverer:
             return
 
         try:
-            recover(statements_from_download(downloaded_data), cursor)
+            recover(statements_from_snapshot(downloaded_data), cursor)
         except Exception as e:
             self._set_state(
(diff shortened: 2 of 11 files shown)


src/_zkapauthorizer/model.py
```diff
@@ -278,11 +278,14 @@ def from_node_config(cls, node_config, now, connect=None):
...
-    @with_cursor
-    def call_if_empty(self, cursor, f: Callable[[Cursor], None]) -> None:
+    @with_cursor_async
+    def call_if_empty(self, cursor, f: Callable[[Cursor], _T]) -> _T:
         """
         Transactionally determine that the database is empty and call the given
         function if it is or raise ``NotEmpty`` if it is not.
+
+        The function may return an ``Awaitable``.  If it does the transaction
+        opened for it will be kept open until the ``Awaitable`` completes.
```
meejah, line 288: I haven't done much "typing in python" but the function signature doesn't say this ... can it?
(Or I guess is there a "async-def Callable" type-signature?)

src/_zkapauthorizer/recover.py
```diff
@@ -187,11 +177,19 @@ async def canned_downloader(set_state: SetState) -> Awaitable[BinaryIO]:
...
-def statements_from_download(data: BinaryIO) -> Iterator[str]:
+def statements_from_snapshot(data: BinaryIO) -> Iterator[str]:
     """
     Read the SQL statements which constitute the replica from a byte string.
     """
-    return data.read().decode("ascii").splitlines()
+    s = data.read()
+    pos = 0
+    while pos < len(s):
+        delim = s.index(b":", pos)
```
meejah, line 187: Not asking for changes .. but this doesn't match by understanding of "netstrings" however it _does_ match what Wikipedia says. (FWIW, my understanding is that you'd interpret the first 2 or 4 or whatever bytes as an int and then read that number of bytes as "the string" ... but now I don't know where that understanding came from ;).

Anyway, I think "bytes" here is at least more-clear than unicode.

src/_zkapauthorizer/resource.py
```diff
@@ -256,23 +257,59 @@ def render_POST(self, request):
             request.setResponseCode(BAD_REQUEST)
             return b"recovery-capability must be a read-only dircap string"
 
+        # The response to this request does not wait for recovery to complete.
+        # Instead, a separate endpoint exposes the progress of the recovery
+        # attempt.  So we're not going to wait for `recovering` to complete.
+        # However, we do have to schedule it in the event loop or it will
+        # never even start.
+        recovering = self._recover(request, self.store, cap)
+        Deferred.fromCoroutine(recovering)
```
meejah, line 266: We're just dropping this `Deferred` .. should it at least have an `addErrback` that does something appropriate? (e.g. at least log the error?) I guess ideally it should tell "something" that the recovery is now never going to happen.

meejah (approved):

src/_zkapauthorizer/model.py
exarkun: Oops.  The "may" here is left-over from when I tried to support both synchronous and asynchronous functions here but I gave up on that.  Now it *must* be an async function.

As far as an "async-def Callable", as far as I know this is written as `Callable[..., Awaitable[...]]`.

Since `f` _must_ be an async function now, the `_T` in the signature should probably be replaced by `Awaitable[_T]`, making it:

```py
def call_if_empty(self, cursor, f: Callable[[Cursor], Awaitable[_T]]) -> Awaitable[_T]:
```

So - `f` is a function that returns an `Awaitable` that completes with a `_T` and `call_if_empty` is a function that returns the same.

How does that look?

src/_zkapauthorizer/recover.py
exarkun: I've usually heard that protocol called "int 16 strings" or "int 32 strings".  Somewhere else I did link to djb's description of the protocol implemented here - https://cr.yp.to/proto/netstrings.txt

Twisted follows the same naming - eg https://docs.twistedmatrix.com/en/stable/api/twisted.protocols.basic.Int16StringReceiver.html and https://docs.twistedmatrix.com/en/stable/api/twisted.protocols.basic.NetstringReceiver.html

PR #341, opened 2022-03-23
remove unused packaging-related files

exarkun:
These files aren't used by anything anymore.


meejah (approved):
Not much context ... so: sure :) I like all-delete PRs

PR #342, opened 2022-03-23
Make recovery work against a real snapshot in the real way

### LINKED ISSUE #333: _zkapauthorizer.recover.recover is incompatible with VoucherStore.call_if_empty

exarkun:
Stacked on #340 

Fixes #333

### DIFF
--- a/src/_zkapauthorizer/tests/test_model.py
+++ b/src/_zkapauthorizer/tests/test_model.py
@@ -56,7 +56,7 @@ from testtools.matchers import (
     Raises,
 )
 from testtools.twistedsupport import failed, succeeded
-from twisted.internet.defer import Deferred
+from twisted.internet.defer import Deferred, succeed
 from twisted.python.runtime import platform
 
 from ..model import (
@@ -91,7 +91,7 @@ from .strategies import (
 _T = TypeVar("T")
 
 
-def fail(cursor):
+async def fail(cursor):
     raise Exception("Should not be called")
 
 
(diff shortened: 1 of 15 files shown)


meejah:
note: I merged in `main` just now

src/_zkapauthorizer/tests/test_model.py
```diff
@@ -1153,3 +1162,45 @@ def test_roundtrip(self, pass_):
...
+        ``StatefulRecoverer.recover``.
+        """
+        store = self.useFixture(
+            ConfiglessMemoryVoucherStore(
+                # Time is not relevant to this test
+                datetime.now,
+            )
+        ).store
+        # XXX Use VoucherStore.snapshot once it is available
+        snapshot_bytes = b"".join(statements_to_snapshot(snapshot(store._connection)))
```
meejah, line 1185: Not really part of this PR I guess, but that reads weirdly .. maybe pointing to a different name for `snapshot()`? (e.g. "`statements()`" if it is read literally, probably, but that's not very descriptive.

E.g. `statements_to_snapshot(snapshot_statements(...))` maybe...?

src/_zkapauthorizer/tests/test_model.py
```diff
@@ -1153,3 +1162,45 @@ def test_roundtrip(self, pass_):
...
+        # It is not enough for StatefulRecoverer.recover to succeed, though.
+        # StatefulRecoverer must also end up in the succeeded state.
+        self.assertThat(
+            recoverer.state(),
+            Equals(
+                RecoveryState(
+                    stage=RecoveryStages.succeeded,
+                ),
+            ),
+        )
```
meejah, line 1206: Feels like this wants more (or maybe even a different test) about whether the database itself ends up in the right state, too? That is, a `recover` that just set the status to `succeeded` would pass this.

meejah (approved):

src/_zkapauthorizer/tests/test_model.py
exarkun: I think this is covered by the `StatefulRecoverer` tests - `_zkapauthorizer.tests.test_recover.StatefulRecovererTests` (particular `test_state_recovered`) - which goes into more detail for the actual recovery code.

exarkun: I added some more words to the class docstring to try to describe its intended limited scope.

PR #343, opened 2022-03-23
Split out replication implementation

exarkun:
This pulls some pieces of the *replication* implementation (ie, the part that creates and maintains a replica) out of the `_zkapauthorizer.recover` module and puts them into a new `_zkapauthorizer.replicate` module.  It leaves the former module for pieces that will download and consume a replica.


meejah:
Aside: _Is_ codecov technically correct, or what? AFAICT you've just moved some (previously uncovered) code to a new spot, but somehow codecov thinks that makes project-coverage drop?

meejah (approved):
:+1: just moved code

exarkun:
> Aside: _Is_ codecov technically correct, or what? AFAICT you've just moved some (previously uncovered) code to a new spot, but somehow codecov thinks that makes project-coverage drop?

I've seen these ~5kloc swings in coverage on a few PRs recently ... in both directions ... so I guess it is just spurious :/  Whether it is a bug in codecov or a problem with our coverage collection/reporting... I dunno.

exarkun:
> > Aside: _Is_ codecov technically correct, or what? AFAICT you've just moved some (previously uncovered) code to a new spot, but somehow codecov thinks that makes project-coverage drop?
> 
> I've seen these ~5kloc swings in coverage on a few PRs recently ... in both directions ... so I guess it is just spurious :/ Whether it is a bug in codecov or a problem with our coverage collection/reporting... I dunno.

Oh, though this is different, it's just a few lines different in this case ... but I also have no explanation for that.

PR #344, opened 2022-03-23
337.upload snapshots

### LINKED ISSUE #337: Upload a new snapshot to remote storage

meejah:
Fixes #337 (non-fork version of PR)


src/_zkapauthorizer/tests/test_recover.py
```diff
@@ -337,35 +336,28 @@ def setUp(self):
...
-                "snapshot.sql",
-                snapshot_cap_str,
-            )
+
+        # use the uploader to push some replica data
+        upload = get_tahoe_lafs_direntry_uploader(
+            tahoeclient,
+            replica_dir_cap_str,
         )
+        yield Deferred.fromCoroutine(upload(BytesIO(b"snapshot data")))
```
exarkun, line 358:
```suggestion
        expected = b"snapshot data"
        yield Deferred.fromCoroutine(upload(BytesIO(expected)))
```

src/_zkapauthorizer/tests/test_recover.py
```diff
@@ -374,7 +366,7 @@ def test_get_downloader(self):
         )
         self.assertThat(
             downloaded_snapshot_path.getContent(),
-            Equals(snapshot_path.getContent()),
+            Equals(b"snapshot data"),
```
exarkun, line 369:
```suggestion
            Equals(expected),
```

src/_zkapauthorizer/tests/__init__.py
```diff
@@ -51,7 +51,7 @@ def _configure_hypothesis():
 
     settings.register_profile(
         "fast",
-        max_examples=2,
+        max_examples=3,
```
exarkun, line 54: intentional?

src/_zkapauthorizer/tahoe.py
```diff
@@ -300,6 +332,9 @@ def download(self, outpath, cap, child_path):
     def upload(self, inpath):
         return upload(self.client, inpath, self._api_root)
 
+    def upload_bytes(self, data):
+        return upload_bytes(self.client, data, self._api_root)
```
exarkun, line 336: This should have a new test or two added to `_zkapauthorizer.tests.test_tahoe`, probably on `UploadDownloadTestsMixin` - and then the test double will need an implementation too.

src/_zkapauthorizer/tahoe.py
```diff
@@ -125,19 +126,21 @@ class NotWriteableError(Exception):
 
 
 @async_retry(_common_tahoe_errors)
-async def upload(
-    client: HTTPClient, inpath: FilePath, api_root: DecodedURL
+async def upload_bytes(
+    client: HTTPClient,
+    data: BytesIO,
```
exarkun, line 131: I think this should be `typing.BinaryIO` rather than `BytesIO` (and I think I incorrectly used the latter a few places in this codebase and haven't fixed them yet).

src/_zkapauthorizer/recover.py
```diff
@@ -249,10 +252,49 @@ async def tahoe_lafs_downloader(
     snapshot_path = client.get_private_path("snapshot.sql")
 
     set_state(RecoveryState(stage=RecoveryStages.downloading))
+    # snapshot-<seqnum>.sql or something?
+
+    # maybe just: snapshot.sql (and we overwrite the thing in the
+    # mutable .. because same failure mode either way?)
```
exarkun, line 258: Can you turn this into some edits to the design doc?

src/_zkapauthorizer/recover.py
```diff
@@ -83,6 +83,9 @@ def marshal(self) -> dict[str, Optional[str]]:
 # An object which can retrieve remote ZKAPAuthorizer state.
 Downloader = Callable[[SetState], Awaitable[BinaryIO]]
 
+# function which can set remote ZKAPAuthorizer state.
+Uploader = Callable[[SetState], Awaitable[None]]
```
exarkun, line 87: Can you put `Uploader` and the related upload functionality below into the new `replicate.py`?

src/_zkapauthorizer/_plugin.py
```diff
@@ -205,6 +207,7 @@ def get_downloader(cap):
             node_config,
             store=self._get_store(node_config),
             get_downloader=get_downloader,
+            ##            get_uploader=make_fail_uploader(work_in_progress_error),
```
exarkun, line 210: I'm not sure the web resource will need the uploader at all.  The "turn on recovery" already gets the `setup_replication` callable.  If upload is necessary, maybe it will be hidden behind that function (but also that function currently only writes out some config, does not directly do any uploads itself).

exarkun (changes requested):
Thanks.  Some comments inline.

src/_zkapauthorizer/tests/__init__.py
meejah: oh, oops .. was trying to see how much difference changing that makes (answer: unknowable, at these sizes).

src/_zkapauthorizer/tahoe.py
meejah: Does it, though? It's already tested via `upload()` -- since it's basically just the guts of it. (I guess I'm "hoping" that we only need one of these, not both...)

exarkun: It looked to me like the test double didn't have an implementation of `upload_bytes` at all - so any test code that uses the test double will break if it tries to call that method?

Also I don't think I object to the path-based upload going away; the new one seems better.

src/_zkapauthorizer/replicate.py
```diff
@@ -78,3 +79,37 @@ async def setup_tahoe_lafs_replication(client: Tahoe) -> Awaitable[str]:
...
     # Return the read-cap
     return rocap
+
+
+async def tahoe_lafs_uploader(
+    client: Tahoe,
+    recovery_cap: str,
+    snapshot_data: BinaryIO,
+    entry_name: str,
+) -> Awaitable[None]:
```
exarkun, line 89:
```suggestion
) -> None:
```

src/_zkapauthorizer/replicate.py
```diff
@@ -78,3 +79,37 @@ async def setup_tahoe_lafs_replication(client: Tahoe) -> Awaitable[str]:
...
+    """
+    Bind a Tahoe client to a mutable directory in a callable that will
+    upload a BytesIO and link it into the mutable directory under the
+    given name.
+
+    :return: A callable that uploads a BytesIO as the latest replica
+        snapshot
+    """
+
+    async def upload(data: BinaryIO) -> Awaitable[None]:
```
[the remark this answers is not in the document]
exarkun: (Also I see that I recently added some other `async def` functions annotated to return `Awaitable`, probably underscores the need for tool-assisted checking)

exarkun (approved):
Thanks!  Just a couple more minor inline comments/suggestions.  Please address those and then merge.

src/_zkapauthorizer/replicate.py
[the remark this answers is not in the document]
meejah: Really? That seems a little counter-intuitive .. the return from an `async-def` is a `coroutine`, which is Awaitable (I would assume)?
...but also I try to ignore mypy ;) so :shrug:

exarkun:
Regarding the macOS CI failure:

```
  File "/Users/distiller/project/venv/lib/python3.9/site-packages/twisted/internet/defer.py", line 1656, in _inlineCallbacks
    result = current_context.run(
  File "/Users/distiller/project/venv/lib/python3.9/site-packages/twisted/python/failure.py", line 489, in throwExceptionIntoGenerator
    return g.throw(self.type, self.value, self.tb)
  File "/Users/distiller/project/venv/lib/python3.9/site-packages/_zkapauthorizer/tests/test_tahoe.py", line 156, in test_not_directory
    cap = yield Deferred.fromCoroutine(client.upload(BytesIO(content)))
  File "/Users/distiller/project/venv/lib/python3.9/site-packages/twisted/internet/defer.py", line 1660, in _inlineCallbacks
    result = current_context.run(gen.send, result)
  File "/Users/distiller/project/venv/lib/python3.9/site-packages/_zkapauthorizer/tahoe.py", line 41, in decorated
    result = await f(*a, **kw)
  File "/Users/distiller/project/venv/lib/python3.9/site-packages/_zkapauthorizer/tahoe.py", line 155, in upload_bytes
    resp = await client.put(uri, data)
  File "/Users/distiller/project/venv/lib/python3.9/site-packages/treq/client.py", line 168, in put
    return self.request('PUT', url, data=data, **kwargs)
  File "/Users/distiller/project/venv/lib/python3.9/site-packages/treq/client.py", line 244, in request
    bodyProducer, contentType = self._request_body(data, files, json,
  File "/Users/distiller/project/venv/lib/python3.9/site-packages/treq/client.py", line 392, in _request_body
    self._data_to_body_producer(data),
  File "/Users/distiller/project/venv/lib/python3.9/site-packages/twisted/python/components.py", line 105, in _hook
    return factory(ob)
  File "/Users/distiller/project/venv/lib/python3.9/site-packages/treq/client.py", line 526, in _from_file
    return FileBodyProducer(orig_file)
  File "/Users/distiller/project/venv/lib/python3.9/site-packages/twisted/web/client.py", line 536, in __init__
    self.length = self._determineLength(inputFile)
  File "/Users/distiller/project/venv/lib/python3.9/site-packages/twisted/web/client.py", line 549, in _determineLength
    originalPosition = tell()
ValueError: I/O operation on closed file.
```

from `_zkapauthorizer.tests.test_tahoe.UploadDownloadIntegrationTests.test_not_directory` I think this may be a real failure caused by the interaction between the retry implementation and the `BytesIO` object.  In the FilePath-based implementation the file would be re-opened for each retry.  Now there is a single `BinaryIO` re-used for each retry... and it looks like the HTTP client does not handle that object in a way that makes it safe to try again :/

The test suite is lacking - it does not deterministically exercise the retry behavior so the macOS CI job just got unlucky and encountered happened to need to try again on that run, where it failed.

meejah:
I tried a different API that's retry-friendly.

exarkun (approved):

PR #345, opened 2022-03-28
Replication layer between SQLite3 and VoucherStore

### LINKED ISSUE #328: Have a replication-capable (snapshot only) database layer
Related to #235 it is necessary to have cooperation at the database layer to take snapshots and capture the "event stream". 

This issue is just about capturing the snapshot and making it available so that a replication service could upload it to remote storage if necessary.  The layer will be amenable to future expansion to capture and expose the event stream as well.

exarkun:
Fixes #328

### DIFF
--- a/src/_zkapauthorizer/config.py
+++ b/src/_zkapauthorizer/config.py
@@ -20,6 +20,7 @@ from datetime import timedelta
 from typing import Any, Optional
 
 from allmydata.node import _Config
+from attrs import define
 from hyperlink import DecodedURL
 from twisted.python.filepath import FilePath
 
@@ -30,17 +31,23 @@ from . import NAME
 REPLICA_RWCAP_BASENAME = NAME + ".replica-rwcap"
 
 
-class _EmptyConfig(object):
+@define
+class EmptyConfig:
     """
     Weakly pretend to be a Tahoe-LAFS configuration object with no
     configuration.
     """
 
+    _basedir: FilePath = FilePath(".")
+
     def get_config(self, section, option, default=object(), boolean=False):
         return default
 
+    def get_private_path(self, name):
+        return self._basedir.child("private").child(name).path
 
-empty_config = _EmptyConfig()
+
+empty_config = EmptyConfig()
 
 
 def read_node_url(config: _Config) -> DecodedURL:
--- a/src/_zkapauthorizer/model.py
+++ b/src/_zkapauthorizer/model.py
@@ -140,10 +140,9 @@ def open_and_initialize(path: FilePath, connect: _ConnectFunction) -> Connection
         run_schema_upgrades(schema_upgrades, cursor)
 
         # Create some tables that only exist (along with their contents) for
-        # this connection.  These are outside of the schema because they are
-        # not persistent.  We can change them any time we like without
-        # worrying about upgrade logic because we re-create them on every
-        # connection.
+        # this connection.  These are outside of the schema because they are not
+        # persistent.  We can change them any time we like without worrying about
+        # upgrade logic because we re-create them on every connection.
         cursor.execute(
             """
             -- Track tokens in use by the process holding this connection.
@@ -160,9 +159,9 @@ def open_and_initialize(path: FilePath, connect: _ConnectFunction) -> Connection
         )
         cursor.execute(
             """
-            -- Track tokens that we want to remove from the database.  Mainly
-            -- just works around the awkward DB-API interface for dealing with
-            -- deleting many rows.
+            -- Track tokens that we want to remove from the database.  Mainly just
+            -- works around the awkward DB-API interface for dealing with deleting
+            -- many rows.
             CREATE TEMPORARY TABLE [to-discard] (
                 [unblinded-token] text
             )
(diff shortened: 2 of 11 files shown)


src/_zkapauthorizer/config.py
```diff
@@ -30,17 +31,23 @@
...
     configuration.
     """
 
+    _basedir: FilePath = FilePath(".")
+
     def get_config(self, section, option, default=object(), boolean=False):
         return default
 
+    def get_private_path(self, name):
+        return self._basedir.child("private").child(name).path
```
exarkun, line 47: This supports the change to the `ConfiglessMemoryVoucherStore` fixture which now uses `VoucherStore.from_node_config(...)` instead of `VoucherStore(...)`.  `from_node_config` uses `get_private_path` to find the SQLite3 database file in the filesystem.

src/_zkapauthorizer/model.py
```diff
@@ -93,20 +96,21 @@ class NotEnoughTokens(Exception):
 # all.  The schema inside the database is versioned by yet another mechanism.
 CONFIG_DB_NAME = "privatestorageio-zkapauthz-v1.sqlite3"
 
+_ConnectParamSpec = [str, int, bool, str, bool]
+_ConnectFunction = Callable[_ConnectParamSpec, Connection]
 
-def open_and_initialize(path, connect=None):
+
+def open_and_initialize(path: FilePath, connect: _ConnectFunction) -> Connection:
```
exarkun, line 103: `connect` is still optional at the higher level API `VoucherStore.from_node_config` but there seemed to be no reason to provide a default for this low-level API.

src/_zkapauthorizer/model.py
```diff
@@ -236,9 +240,6 @@ class VoucherStore(object):
     """
     This class implements persistence for vouchers.
 
-    :ivar allmydata.node._Config node_config: The Tahoe-LAFS node configuration object for
-        the node that owns the persisted vouchers.
-
```
exarkun, line 241: It hasn't actually had this attribute for a while.

src/_zkapauthorizer/tests/test_model.py
```diff
@@ -245,20 +245,27 @@ def test_empty(self):
...
-            self.store_fixture.store._connection.execute("SELECT * FROM [it_ran]")
+
+        async def check_side_effect(cursor):
+            rows = cursor.execute("SELECT * FROM [it_ran]")
+            rows = cursor.fetchall()
+            return rows
+
+        self.assertThat(
+            Deferred.fromCoroutine(store.call_if_empty(check_side_effect)),
+            succeeded(Equals([(1,)])),
```
exarkun, line 267: This hunk just let's the test avoid touching `store._connection` directory.

src/_zkapauthorizer/replicate.py
```diff
@@ -78,3 +117,133 @@ async def setup_tahoe_lafs_replication(client: Tahoe) -> Awaitable[str]:
...
+        return self._cursor.fetchmany(n)
+
+    def fetchone(self):
+        return self._cursor.fetchone()
+
+    def executemany(self, statement, rows):
+        self._cursor.executemany(statement, rows)
+
+
+def netstring(bs: bytes) -> bytes:
```
exarkun, line 206: `netstring` and `statements_to_snapshot` just moved here from `test_recover.py`.  `connection_to_statements` is the old `snapshot` but with newlines removed because we do netstrings now.

src/_zkapauthorizer/config.py
meejah: This smells like "asking for the ingredients" instead of asking for the cake -- can `VoucherStore` ask for the database itself instead?

I guess that means: wrap up "the path name" -- if indeed it's relevant at all -- into the `connect` function that's already passed through to `open_and_initialize`..? Indeed `memory_connect` already just throws the "path" argument away...

exarkun: I like this idea.  I *think* I see how the `node_config` parts of `VoucherStore.from_node_config` could be pulled up to `_plugin.py` (or over to `config.py` and called from `_plugin.py`).

exarkun: I have done some refactoring and I think the result is better than what's here now - but I didn't manage to get rid of this new `get_private_path` since _something_ still needs to compute a database filesystem path relative to the Tahoe-LAFS node directory, and for a while that's going to have to use the config object's `get_private_path`.

I didn't push those refactorings to this PR because they mostly just make it larger, I'll put them up separately though.

meejah (approved):

src/_zkapauthorizer/replicate.py
```diff
@@ -142,6 +179,135 @@ async def setup_tahoe_lafs_replication(client: Tahoe) -> Awaitable[str]:
...
+    """
+    Encode a single string as a netstring.
+
+    :see: http://cr.yp.to/proto/netstrings.txt
```
hacklschorsch, line 270: 👍

hacklschorsch (commented):
I like this a lot!

PR #348, opened 2022-03-29
Make `_zkapauthorizer` "type check"

exarkun:
This adds a mypy configuration and makes some changes to allow mypy to run cleanly over the codebase.

It also sets up a CI job to run mypy.


exarkun:
unlandable mess, will clean up and try again elsewhere

PR #350, opened 2022-03-31
WIP: 349.event stream

### LINKED ISSUE #349: Implement an "event stream" object
Inside a replica, the event-stream represents SQL statements that change the database and were run _after_ the last snapshot was created.

These are collected and put into the local database (I suppose: being careful to not also try to event-stream _those_ statements themselves??)

Periodically, we upload them.

When a new snapshot is created, all Events contained in it are pruned locally.

Tasks for this ticket: 
- add database table for event-streams
- serialize Events into the database
- write an "event stream" abstraction. Must be immutable (other parts of the code may already assume this)
- serialize it to bytes

meejah:
Fixes #349 

(Push early, push often, right? ... Still very much WIP but someone might want to look early...)

### DIFF
--- a/src/_zkapauthorizer/model.py
+++ b/src/_zkapauthorizer/model.py
@@ -39,6 +39,7 @@ from .storage_common import (
     required_passes,
 )
 from .validators import greater_than, has_length, is_base64_encoded
+from .replicate import EventStream, Change
 
 _T = TypeVar("T")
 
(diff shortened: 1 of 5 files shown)


meejah:
(clicked "create" too fast ... and apparently if you forget to select "Draft" then too bad for you)

meejah:
Ah thanks ... that's obscure ;)

PR #351, opened 2022-03-31
349.event stream

### LINKED ISSUE #349: Implement an "event stream" object

meejah:
fixes #349

### DIFF
--- a/src/_zkapauthorizer/model.py
+++ b/src/_zkapauthorizer/model.py
@@ -39,6 +39,7 @@ from .storage_common import (
     required_passes,
 )
 from .validators import greater_than, has_length, is_base64_encoded
+from .replicate import EventStream, Change
 
 _T = TypeVar("T")
 
@@ -786,9 +787,9 @@ class VoucherStore(object):
         """
         cursor.execute(
             """
-            INSERT INTO [event-stream] VALUES (?)
+            INSERT INTO [event-stream]([statement]) VALUES (?)
             """,
-            sql_statement
+            (sql_statement, )
         )
 
     @with_cursor
@@ -796,16 +797,17 @@ class VoucherStore(object):
         """
         Return all events currently in our event-log.
         """
-        cursor.execute(
+        rows = cursor.execute(
             """
             SELECT [sequence-number], [statement]
             FROM [event-stream]
             """
-        )
+        ).fetchall()
+
         return EventStream(
-            events=[
+            changes=[
                 Change(stmt)
-                for seq, stmt in row.fetchall()
+                for seq, stmt in rows
             ]
         )
 
--- a/src/_zkapauthorizer/model.py
+++ b/src/_zkapauthorizer/model.py
@@ -803,7 +803,7 @@ class VoucherStore(object):
             """
         ).fetchall()
 
-        return EventStream(changes=[Change(stmt, seq) for seq, stmt in rows])
+        return EventStream(changes=[Change(seq, stmt) for seq, stmt in rows])
 
 
 @implementer(ILeaseMaintenanceObserver)
--- a/src/_zkapauthorizer/model.py
+++ b/src/_zkapauthorizer/model.py
@@ -32,6 +32,7 @@ from zope.interface import Interface, implementer
 
 from ._base64 import urlsafe_b64decode
 from ._json import dumps_utf8
+from .replicate import Change, EventStream
 from .schema import get_schema_upgrades, get_schema_version, run_schema_upgrades
 from .storage_common import (
     get_configured_pass_value,
@@ -39,7 +40,6 @@ from .storage_common import (
     required_passes,
 )
 from .validators import greater_than, has_length, is_base64_encoded
-from .replicate import EventStream, Change
 
 _T = TypeVar("T")
 
@@ -789,7 +789,7 @@ class VoucherStore(object):
             """
             INSERT INTO [event-stream]([statement]) VALUES (?)
             """,
-            (sql_statement, )
+            (sql_statement,),
         )
 
     @with_cursor
@@ -804,12 +804,7 @@ class VoucherStore(object):
             """
         ).fetchall()
 
-        return EventStream(
-            changes=[
-                Change(stmt)
-                for seq, stmt in rows
-            ]
-        )
+        return EventStream(changes=[Change(stmt) for seq, stmt in rows])
 
 
 @implementer(ILeaseMaintenanceObserver)
--- a/src/_zkapauthorizer/model.py
+++ b/src/_zkapauthorizer/model.py
@@ -778,6 +778,37 @@ class VoucherStore(object):
             parse_datetime(finished, delimiter=" "),
         )
 
+    # XXX do we want "str" or "bytes" here?
+    @with_cursor
+    def add_event(self, cursor, sql_statement: str):
+        """
+        Add a new event to the event-log.
+        """
+        cursor.execute(
+            """
+            INSERT INTO [event-stream] VALUES (?)
+            """,
+            sql_statement
+        )
+
+    @with_cursor
+    def get_events(self, cursor):
+        """
+        Return all events currently in our event-log.
+        """
+        cursor.execute(
+            """
+            SELECT [sequence-number], [statement]
+            FROM [event-stream]
+            """
+        )
+        return EventStream(
+            events=[
+                Change(stmt)
+                for seq, stmt in row.fetchall()
+            ]
+        )
+
 
 @implementer(ILeaseMaintenanceObserver)
 @attr.s
--- a/src/_zkapauthorizer/model.py
+++ b/src/_zkapauthorizer/model.py
@@ -779,11 +779,10 @@ class VoucherStore(object):
             parse_datetime(finished, delimiter=" "),
         )
 
-    # XXX do we want "str" or "bytes" here?
     @with_cursor
     def add_event(self, cursor, sql_statement: str):
         """
-        Add a new event to the event-log.
+        Add a new change to the event-log.
         """
         cursor.execute(
             """
@@ -804,7 +803,7 @@ class VoucherStore(object):
             """
         ).fetchall()
 
-        return EventStream(changes=[Change(stmt) for seq, stmt in rows])
+        return EventStream(changes=[Change(stmt, seq) for seq, stmt in rows])
 
 
 @implementer(ILeaseMaintenanceObserver)
--- a/src/_zkapauthorizer/replicate.py
+++ b/src/_zkapauthorizer/replicate.py
@@ -38,8 +38,8 @@ class Change:
     Represent an item in a replication event stream
     """
 
-    statement: str  # the SQL statement string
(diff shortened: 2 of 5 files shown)


src/_zkapauthorizer/recover.py
```diff
@@ -190,6 +190,36 @@ def statements_from_snapshot(data: BinaryIO) -> Iterator[str]:
...
+def statements_to_snapshot(statements: Iterator[str]) -> Iterator[bytes]:
+    """
+    Take a snapshot of the database reachable via the given connection.
+    """
+    for statement in statements:
+        # Use netstrings to frame each statement.  Statements can have
+        # embedded newlines (and CREATE TABLE statements especially tend to).
+        yield netstring(statement.encode("utf-8"))
+
+
```
meejah, line 222: above i'm just moving the code from a test module

meejah:
Local tools say 98% coverage. Can't explain why codecov thinks 90%.

I can't tell for _sure_ if I'm running it the same way as in CI, but inspecting the logs the command-line is the same. (I also tried with an explicit `--branch` even though the config says that, but the result is the same).

src/_zkapauthorizer/model.py
```diff
@@ -778,6 +779,32 @@ def get_latest_lease_maintenance_activity(self, cursor):
             parse_datetime(finished, delimiter=" "),
         )
 
+    @with_cursor
```
exarkun, line 782: It might be that the caller ends up being responsible for this transaction management because these inserts will need to run in the same transaction as the statements themselves come from.  I don't think there's a reason to change it now rather than later, just commenting to share my hunch about where this will have to go.

src/_zkapauthorizer/model.py
```diff
@@ -778,6 +779,32 @@ def get_latest_lease_maintenance_activity(self, cursor):
...
+        Return all events currently in our event-log.
+        """
+        rows = cursor.execute(
+            """
+            SELECT [sequence-number], [statement]
+            FROM [event-stream]
+            """
+        ).fetchall()
+
+        return EventStream(changes=tuple(Change(seq, stmt) for seq, stmt in rows))
```
exarkun, line 806: Hmm.  Feels weird but ... an immutable container that preserves order... yea, I guess that's a tuple.

src/_zkapauthorizer/replicate.py
```diff
@@ -23,14 +23,77 @@
...
+class EventStream:
+    """
+    A series of database operations represented as `Change` instances.
+    """
+
+    changes: tuple[Change]
+
+    def highest_sequence(self):
+        """
+        :return int: the highest sequence number in this EventStream (or
```
exarkun, line 57: I guess `Optional[int]`.  Also I've been slowly moving type info from the docstrings into annotations, so maybe this could move similarly.

src/_zkapauthorizer/replicate.py
```diff
@@ -23,14 +23,77 @@
...
+            None if there are no events)
+        """
+        if not self.changes:
+            return None
+        return max(change.sequence for change in self.changes)
+
+    # XXX or should we pass in a writable stream to use instead?
```
exarkun, line 64: I think the only way we could actually realize a benefit is if this could stream right into treq ... so, an `IProducer`?  which I think is still kind of tedious and fiddly.  So probably an optimization for later.

src/_zkapauthorizer/replicate.py
```diff
@@ -23,14 +23,77 @@
...
+                {
+                    "events": tuple(
+                        (event.sequence, event.statement.encode("utf8"))
+                        for event in self.changes
+                    )
+                }
+            )
+        )
+
+    # XXX versioning? or do we handle that higher up?
```
exarkun, line 80: We could either put the version inside the stream or in metadata we find about the stream (the filename or metadata in the direntry).

It would probably be essentially trivial to put a `.v1` on the end of the filename and then declare versioning solved until we need a version 2.

src/_zkapauthorizer/schema.py
```diff
@@ -208,4 +208,18 @@ def run_schema_upgrades(upgrades, cursor):
         ALTER TABLE [unblinded-tokens] ADD COLUMN [redemption-group] integer DEFAULT 1
         """,
     ],
+    6: [
+        """
+        -- track the "event-stream" which are a series of SQL statements
+        -- that modify the database (except statements which modify this table?)
```
exarkun, line 214: the exception here is correct.  the capture layer will have to know about this table and exclude it from capturing, otherwise I assume you have the very predictable infinite recursion.

src/_zkapauthorizer/sql.py
```diff
@@ -71,10 +71,33 @@ def statement(self):
...
     def arguments(self):
         return self.fields
 
 
+def quote_sql_value(cursor, value):
+    if isinstance(value, int):
+        return str(value)
+    if isinstance(value, float):
+        return str(value)
+    if isinstance(value, type(None)):
```
exarkun, line 94: could just be `is None`

src/_zkapauthorizer/tests/test_model.py
```diff
@@ -856,6 +865,72 @@ def test_lease_maintenance_activity(self, get_config, now, activity):
...
+        """
+        Various kinds of SQL statements can be serialized into and out of
+        the event-stream.
+        """
+        # no BLOBs
+        assume(
+            not any(
+                column[1].affinity == StorageAffinity.BLOB for column in table.columns
+            )
+        )
```
exarkun, line 893: For what it's worth, SQLite3 will also quote these for us with `quote()` so we could support this if we wanted.  I don't think we have any blobs in the schema right now but we could be storing a slightly more compact representation of the cryptographic tokens if we did.

Since we don't need blobs now, change here is certainly optional - but I think blob support in `quote_sql_value` is fewer lines than this assumption.

src/_zkapauthorizer/tests/test_model.py
```diff
@@ -856,6 +865,72 @@ def test_lease_maintenance_activity(self, get_config, now, activity):
...
+
+        events = store.get_events()
+        self.assertThat(
+            events,
+            AfterPreprocessing(
+                lambda eventstream: [
+                    (change.sequence, change.statement)
+                    for change in eventstream.changes
+                ],
+                Equals(sql_statements),
```
exarkun, line 920: Does `Change` not have a good `__eq__`?  Why unpack it like this instead of comparing the list of Changes directly?

src/_zkapauthorizer/replicate.py
```diff
@@ -23,14 +23,77 @@
...
+
+@frozen
+class EventStream:
+    """
+    A series of database operations represented as `Change` instances.
+    """
+
+    changes: tuple[Change]
+
+    def highest_sequence(self):
```
exarkun, line 55: The design doc says that event stream objects have the lowest sequence number they contain in their metadata.  Maybe all of the logic works out with highest too?  Maybe the design doc could be clearer about this logic, too.

Specifically, the words in the design doc are:

> The sequence number of the first statement it includes is added as metadata for that object in the containing directory.

Please make the implementation and doc consistent (change whichever you think it is right to change).

exarkun (approved):
Thanks!  Looks about right to me.  Some comments inline, nothing major I think.  One other thing is that some new APIs have no docstrings - if you could add those, that'd be great.

Then go ahead and merge.

src/_zkapauthorizer/replicate.py
meejah: `treq` can take file-likes too, so anything from the `io.*` family of ... interfaces? base-classes? was more along the lines I was thinking (although I didn't go far enough to actually-try). But, yeah, could change the API later, it's just internal.

src/_zkapauthorizer/replicate.py
meejah: In magic-folders, the versioning debate ended with "we should put something in the Tahoe directory" and that something was an `@metadata` file that is JSON shaped like `{"version": 1}` ... (the `@` works out well there because `@<text>` is otherwise illegal in a magic-folder name _and_ because that actual JSON is small enough to fit in a LIT cap.

So we could do something similar here and declare the entire backup schmea as "v1" and put a `metadata.json` or something to that effect in the mutable-dir.

(...but what I was actually ref'ing in the above comment is that I saw `v1` in some other namings but I wasn't clear if those are meant to encapsulate "this way of writing to Tahoe" or something else).

src/_zkapauthorizer/tests/test_model.py
meejah: Okay, it wasn't clear to me that `quote()` would also work for BLOBs _and_ I figured we weren't using them etc .. but I can just let it through if you think that's better.

src/_zkapauthorizer/replicate.py
meejah: Ahh, hmmm ... okay. I think why I used "highest" is because the design wants to make decisions like "is it too long since that last one" and "does this snapshot contain everything in event-stream X" and both those want the maximum number. (I guess strictly speaking the second question needs _both_ numbers, but a "snapshot that contains sequence number X" is supposed to wrap-up every event with `sequences < X` right?)

src/_zkapauthorizer/model.py
meejah: okay. should i add a comment? (or we just remove the decorator when .. the other code appears?)

src/_zkapauthorizer/tests/test_model.py
meejah: I suppose I create and store `Change`s in the test, yes ... i think it's unpacked because i started just comparing the statements and added sequences later.

meejah: (The `change` from the data.draw isn't the same as `replicate.Change` ...)

src/_zkapauthorizer/replicate.py
meejah: So, in that prototyping branch, event-streams are linked to the replica as `event-stream-<number>` which is I guess either the highest or lowest number in that event-stream (depending which I change).

If it's the lowest, pruning becomes harder. Let's say I have:

- a new snapshot with highest sequence number of 122
- on-replica event-stream files: `event-stream-0`,  `event-stream-123`

Can we delete `event-stream-0`?

We'd have to download it to be sure .. usually it would be the case with the above that the first file contains events 0 -> 122, but we don't really _know_ that. On the other hand, if that was `event-stream-122` and `event-stream-432` then we can tell just from the filenames/metadata (without downloading) that we can delete only `event-stream-122` right now (because a snapshop with "122" contains all the events to that point).

PR #353, opened 2022-04-01
Simple fixes for mypy errors

exarkun:
This fixes 66 of the 77 mypy errors found by #352 .  It's not stacked for easy of review.


src/_zkapauthorizer/_storage_client.py
```diff
@@ -50,7 +50,7 @@
     int,
     tuple[
         list[
-            tuple[int, int, bytes, bytes],
+            tuple[int, int, bytes],
```
exarkun, line 53: This type used to be correct but the representation inside Tahoe-LAFS changed.  Our implementation was already updated (so it would work) but not the type.

src/_zkapauthorizer/_storage_client.py
```diff
@@ -436,13 +436,13 @@ def slot_testv_and_readv_and_writev(
...
         r_vector: ReadVector,
-    ) -> Deferred:
+    ) -> Generator[Deferred[Any], Any, None]:
         # Read operations are free.
         num_passes = 0
 
         # Convert tw_vectors from the new internal format to the wire format.
         # See https://github.com/tahoe-lafs/tahoe-lafs/pull/1127/files#r716939082
-        tw_vectors = {
+        old_tw_vectors = {
```
exarkun, line 445: mypy doesn't like re-assigning the variable with a new type.

src/_zkapauthorizer/controller.py
```diff
@@ -195,6 +195,9 @@ class IndexedRedeemer(object):
 
     redeemers = attr.ib()
 
+    def tokens_to_passes(self, message, unblinded_tokens):
+        raise NotImplementedError("IndexedRedeemer cannot create passes")
```
exarkun, line 199: This method is part of the interface and mypy doesn't like incomplete implementations.

src/_zkapauthorizer/lease_maintenance.py
```diff
@@ -17,10 +17,12 @@
 refresh leases on all shares reachable from a root.
 """
 
+from __future__ import annotations
```
exarkun, line 20: This lets a classmethod on LeaseMaintenanceConfig say it returns a LeaseMaintenanceConfig.

src/_zkapauthorizer/lease_maintenance.py
```diff
@@ -458,7 +457,7 @@ class LeaseMaintenanceConfig(object):
     min_lease_remaining: timedelta = attr.ib()
 
     @classmethod
-    def from_node_config(cls: Type[_T], node_config: Config) -> _T:
```
exarkun, line 461: Some SO post told me this would let you tell mypy that the method returns an instance of `cls` but mypy still complains so I just gave up on that idea.

src/_zkapauthorizer/model.py
```diff
@@ -43,7 +43,7 @@
 )
 from .validators import greater_than, has_length, is_base64_encoded
 
-_T = TypeVar("T")
+_T = TypeVar("_T")
```
exarkun, line 46: mypy demands that the name given matches the name the TypeVar is bound to.

src/_zkapauthorizer/sql.py
```diff
@@ -62,7 +64,7 @@ class Insert:
 
     table_name: str
     table: Table
-    fields: tuple[...]
+    fields: tuple[SQLType, ...]
```
exarkun, line 67: This means "a tuple of indeterminate length with elements of type SQLType"

src/_zkapauthorizer/tahoe.py
```diff
@@ -51,7 +50,7 @@ async def decorated(*a, **kw) -> Awaitable:
     return retry_decorator
 
 
-def _not_enough_servers(exc: Exception) -> bool:
+def _not_enough_servers(exc: Exception, /) -> bool:
```
exarkun, line 53: This makes `exc` positional-only which is apparently required for this type to match `Callable[[Exception], bool]` - which is what is required by the API this is passed to.

It looks like newer Python might give you a way to describe keyword-argument functions but we'll have to wait for that.

src/_zkapauthorizer/tahoe.py
```diff
@@ -430,18 +450,16 @@ async def download(self, outpath, cap, child_path):
...
-        if isinstance(d, dict):
-            # It is a directory.  Encode it somehow so it fits in the file.
-            # This is not the same encoding as Tahoe-LAFS itself uses for
-            # directories.
-            d = dumps_utf8(d)
-        outpath.setContent(d)
+        if isinstance(d, bytes):
+            outpath.setContent(d)
+        else:
+            raise ValueError(f"Cannot download non-data capability ({cap[:7]})")
```
exarkun, line 462: It was basically wrong to let this API download a dircap before - you can't do that with the real tahoe client (or, you can, but you get data in a totally different format than we were giving out).

src/_zkapauthorizer/tests/_sql_matchers.py
```diff
@@ -86,10 +82,10 @@ def _structured_dump_table(
...
     for rows in iter(lambda: curs.fetchmany(1024), []):
         for row in rows:
-            # We have no representation of the table right now so we'll just
-            # leave it out.  This still gives us a convenient container for
-            # the other values.
-            yield Insert(table_name, None, row)
+            # We don't have a representation of the table so we'll supply a
+            # bogus value.  This still gives us a convenient container for the
+            # other values.
+            yield Insert(table_name, Table([]), row)
```
exarkun, line 88: `None` was not allowed for a parameter declared as `Table`.

src/_zkapauthorizer/tests/matchers.py
```diff
@@ -50,13 +52,22 @@
...
 from ._exception import raises
 from ._float_matchers import matches_float_within_distance
 from ._sql_matchers import equals_database
 
+_T = TypeVar("_T")
+
+
+class Matcher(_Matcher, Generic[_T]):
+    """
+    A generic version of ``_Matcher``.
```
exarkun, line 68: in other words, a type that lets you write `Matcher[foo]` to say "this is a thing that matches on instances of foo".

src/_zkapauthorizer/tests/matchers.py
```diff
@@ -220,7 +233,7 @@ def matches_response(
 
 def matches_spent_passes(
     public_key_hash: bytes, spent_passes: list[Pass]
-) -> GenericAlias(Matcher, (_SpendingData,)):
```
exarkun, line 223: This `GenericAlias` thing was just flat out wrong.

src/_zkapauthorizer/tests/test_tahoe.py
```diff
@@ -204,34 +204,14 @@ class DirectoryTestsMixin:
...
-        of a new, empty directory.
-        """
-        tahoe = self.get_client()
-
-        dir_cap = yield Deferred.fromCoroutine(tahoe.make_directory())
-
-        # If we can download it, consider that success.
-        outpath = FilePath(self.useFixture(TempDir()).join("dir_contents"))
-        yield Deferred.fromCoroutine(tahoe.download(outpath, dir_cap, None))
-        self.assertThat(outpath.getContent(), Not(Equals(b"")))
```
exarkun, line 220: This test broke because I disabled downloading directory caps.  You have to `list_directory` them now - which is just what the next test does.

PR #356, opened 2022-04-05
Remove `VoucherStore.from_node_config` in favor of something better factored

exarkun:
This came from @meejah 's comments about how `VoucherStore`'s reliance on the Tahoe-LAFS config object to find its database seemed less than ideal.

Ultimately this code still uses the Tahoe-LAFS config object to find the database path, but it does so in the glue code in `_plugin.py` instead of the database code in `model.py`.

### DIFF
--- a/src/_zkapauthorizer/tests/test_model.py
+++ b/src/_zkapauthorizer/tests/test_model.py
@@ -848,12 +848,9 @@ class EventStreamTests(TestCase):
         Various kinds of SQL statements can be serialized into and out of
         the event-stream.
         """
-        tempdir = self.useFixture(TempDir())
-        store = VoucherStore.from_node_config(
-            get_config(tempdir.join("node"), "tub.port"),
-            lambda: now,
-            memory_connect,
-        )
+        store = self.useFixture(
+            TemporaryVoucherStore(get_config, lambda: now),
+        ).store
 
         # generate some SQL events
         sql_statements = []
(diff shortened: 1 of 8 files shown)


src/_zkapauthorizer/tests/test_model.py
```diff
@@ -492,74 +484,6 @@ def test_list(self, get_config, now, vouchers, data):
             ),
         )
 
-    @skipIf(platform.isWindows(), "Hard to prevent directory creation on Windows")
```
exarkun, line 495: These tests moved to another module and got simplified a bit.

src/_zkapauthorizer/tests/test_model.py
```diff
@@ -1231,28 +1152,6 @@ def test_not_enough_unblinded_tokens(
         )
 
 
-def store_for_test(testcase, get_config, get_now):
```
exarkun, line 1234: This was an older helper that no code uses anymore.

src/_zkapauthorizer/tests/test_plugin.py
```diff
@@ -108,6 +110,58 @@ def get_rref(interface=None):
...
 
+class OpenStoreTests(TestCase):
+    @skipIf(platform.isWindows(), "Hard to prevent directory creation on Windows")
+    @given(tahoe_configs(), datetimes())
+    def test_uncreateable_store_directory(self, get_config, now):
+        """
+        If the underlying directory in the node configuration cannot be created
+        then ``open_store`` raises ``StoreOpenError``.
+        """
+        nodedir = FilePath(self.useFixture(TempDir()).join("node"))
```
meejah, line 121: This looks like a clue about codecov: it maybe only looks at "windows" coverage? (Since it says this test isn't run

meejah (approved):
LGTM

PR #358, opened 2022-04-12
Add a skeleton replication service, hooked up to other things

### LINKED ISSUE #355: Have a Twisted Service that is running when replication is enabled
This is related to #235.  Replication involves executing certain tasks in response to certain events.  New snapshots and event streams must be uploaded.  Obsolete snapshots and event streams must be unlinked.

Responsibility for coordinating these activities can fall on a service object which arranges for information about the database state to be fed into application logic for performing the necessary uploads and unlinkings.

exarkun:
Fixes #355 

This adds a do-nothing replication service that is automatically started for a VoucherStore when that VoucherStore is opened if and only if it finds replication configuration living alongside that VoucherStore.

This is about half of the logic we want for "make sure replication is happening".  The other half is for the case where the VoucherStore is already open and someone turns on replication.

### DIFF
--- a/src/_zkapauthorizer/_plugin.py
+++ b/src/_zkapauthorizer/_plugin.py
@@ -35,7 +35,7 @@ from attrs import Factory, define, field
 from challenge_bypass_ristretto import PublicKey, SigningKey
 from eliot import start_action
 from prometheus_client import CollectorRegistry, write_to_textfile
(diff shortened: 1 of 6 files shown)


meejah (approved):
(more evidence codecov is only looking at windows runs in this PR...)

PR #362, opened 2022-04-14
Use filesystem lock correctly

### LINKED ISSUE #361: The filesystem lock API is used incorrectly by setup_tahoe_lafs_replication
`FilesystemLock.lock` returns True or False but the code doesn't use it that way.

exarkun:
Fixes #361 

Stacked on #360


hacklschorsch (approved):
I think this is all very nice stuff!

PR #365, opened 2022-04-26
Enforce typechecking

exarkun:
This makes the mypy CI job fail builds if mypy fails.  It also fixes the type errors that have crept in since #357

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -254,7 +254,7 @@ jobs:
       - run:
           name: "Run Type Checks"
           command: |
-            nix-shell --run 'mypy src || true'
+            nix-shell --run 'mypy src'
       - run:
           <<: *PUSH_TO_CACHIX
 
--- a/src/_zkapauthorizer/_types.py
+++ b/src/_zkapauthorizer/_types.py
@@ -17,9 +17,7 @@ Re-usable type definitions for ZKAPAuthorizer.
 """
 
 from datetime import datetime
-from typing import Any, Callable, ContextManager, Optional, Protocol, Union
-
-from .sql import SQLType
+from typing import Callable
 
 # A Tahoe-LAFS capability string
 CapStr = str
--- a/src/_zkapauthorizer/_types.py
+++ b/src/_zkapauthorizer/_types.py
@@ -16,29 +16,12 @@
 Re-usable type definitions for ZKAPAuthorizer.
 """
 
-from sqlite3 import Connection
-from typing import Any, Callable, Protocol
+from datetime import datetime
+from typing import Any, Callable, ContextManager, Optional, Protocol, Union
+
+from .sql import SQLType
 
 # A Tahoe-LAFS capability string
 CapStr = str
 
-GetTime = Callable[[], float]
-
-
-class Connect(Protocol):
-    """
-    Connect to a certain (ie, not parameterized) SQLite3 database.
-    """
-
-    def __call__(
-        self,
-        timeout: int = None,
-        detect_types: bool = None,
-        isolation_level: str = None,
-        check_same_thread: bool = False,
-        factory: Any = None,
-        cached_statements: Any = None,
-    ) -> Connection:
-        """
-        Get a new database connection.
-        """
+GetTime = Callable[[], datetime]
--- a/src/_zkapauthorizer/model.py
+++ b/src/_zkapauthorizer/model.py
@@ -23,13 +23,14 @@ import os
 from datetime import datetime
 from functools import wraps
 from json import loads
-from sqlite3 import Connection, Cursor, OperationalError
+from sqlite3 import Connection as _SQLite3Connection
+from sqlite3 import OperationalError
 from sqlite3 import connect as _connect
-from typing import Awaitable, Callable, TypeVar
+from typing import Awaitable, Callable, Optional, TypeVar
 
 import attr
 from aniso8601 import parse_datetime
-from attr import define, frozen
+from attrs import define, frozen
 from hyperlink import DecodedURL
 from twisted.logger import Logger
 from twisted.python.filepath import FilePath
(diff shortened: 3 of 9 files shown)


.circleci/config.yml
```diff
@@ -246,7 +246,7 @@ jobs:
       - run:
           name: "Run Type Checks"
           command: |
-            nix-shell --run 'mypy src || true'
+            nix-shell --run 'mypy src'
```
exarkun, line 249: This actually does what the PR says.

src/_zkapauthorizer/_types.py
```diff
@@ -16,29 +16,10 @@
...
+from datetime import datetime
+from typing import Callable
 
 # A Tahoe-LAFS capability string
 CapStr = str
 
-GetTime = Callable[[], float]
-
-
-class Connect(Protocol):
```
exarkun, line 28: This is moving to the SQL module since there have to be two of them now.

src/_zkapauthorizer/_types.py
```diff
@@ -16,29 +16,10 @@
...
-        detect_types: bool = None,
-        isolation_level: str = None,
-        check_same_thread: bool = False,
-        factory: Any = None,
-        cached_statements: Any = None,
-    ) -> Connection:
-        """
-        Get a new database connection.
-        """
+GetTime = Callable[[], datetime]
```
exarkun, line 25: The previous return type, float, was an error.

src/_zkapauthorizer/model.py
```diff
@@ -88,22 +90,7 @@ class NotEnoughTokens(Exception):
     """
 
 
-def open_and_initialize(connect: Connect) -> Connection:
```
exarkun, line 91: This was unused.

src/_zkapauthorizer/model.py
```diff
@@ -129,7 +116,7 @@ def initialize_database(conn: Connection) -> None:
         # be enabled on a per-connection basis.  This is a helpful feature to
         # ensure consistency so we want it enforced and we use it in our
         # schema.
-        cursor.execute("PRAGMA foreign_keys = ON")
+        cursor.execute("PRAGMA foreign_keys = ON", ())
```
exarkun, line 119: mypy doesn't have variadic functions so we have to pass an empty argument tuple to execute now.

src/_zkapauthorizer/model.py
```diff
@@ -296,7 +287,7 @@ def snapshot(self) -> bytes:
         Create and return a consistent, self-contained snapshot of the underlying
         database state.
         """
-        return self._connection.snapshot()
+        return snapshot(self._connection)
```
exarkun, line 290: since it doesn't make sense for the abstract Connection protocol to have a snapshot method (the real sqlite3 Connection doesn't) I made this method into a function.  I could have annotated `self._connection` as `_ReplicationCapableConnection` instead but I didn't want to depend on that type in `model.py`, and also I got it working this way first and didn't want to try it another way.

src/_zkapauthorizer/replicate.py
```diff
@@ -237,24 +239,33 @@ def enable_replication(self) -> None:
         """
         self._replicating = True
 
-    def snapshot(self) -> bytes:
+    def iterdump(self) -> Iterable[str]:
```
exarkun, line 242: Proxy `iterdump` so the stand-alone `snapshot` function can work on this kind of connection.

src/_zkapauthorizer/sql.py
```diff
@@ -5,15 +5,98 @@
...
+from sqlite3 import Connection as _SQLite3Connection
+from typing import Any, ContextManager, Iterable, Optional, Protocol, Union
 
-from attrs import define
+from attrs import frozen
 
 SQLType = Union[int, float, str, bytes, None]
 
 
+class AbstractCursor(Protocol):
```
exarkun, line 19: codify Cursor and Connection protocols here so we can frequently refer to the abstract protocol instead of a concrete type

exarkun: these are constrained by the actual `sqlite3.Connection` and `sqlite3.Cursor` types because we need them to unify with this protocol.

src/_zkapauthorizer/sql.py
```diff
@@ -5,15 +5,98 @@
...
+        exc_tb: Optional[Any],
+    ) -> None:
+        ...
+
+
+Connection = AbstractConnection
+Cursor = AbstractCursor
+
+
+class UnboundConnect(Protocol):
```
exarkun, line 61: It turns out we needed annotations for both "connect that accepts a path" and "connect that is already partially applied to a path" so there are two of these now.

src/_zkapauthorizer/sql.py
```diff
@@ -27,7 +110,7 @@ class StorageAffinity(Enum):
     NUMERIC = auto()
 
 
-@define(frozen=True)
+@frozen
```
exarkun, line 113: incidental, I was trying to fix a problem where hypothesis suddenly became unable to automatically build Columns.  this change didn't help but it seemed worth keeping.

src/_zkapauthorizer/tests/test_plugin.py
```diff
@@ -190,14 +194,20 @@ def _replication_enabled_connection_test(self, now: datetime, enabled: bool):
...
 
-        self.assertThat(
-            # Right now enabling replication does absolutely nothing except
-            # flip this flag, so there's no other behavior to observe in this
-            # test.  Maybe later when there's more replication implementation,
-            # we can assert something more meaningful.
-            store._connection._replicating,
-            Equals(enabled),
-        )
+        if isinstance(store._connection, _ReplicationCapableConnection):
```
exarkun, line 197: if we check the type of `store._connection` then mypy allows us to touch attributes of the concrete type.

hacklschorsch (approved):
LGTM!

OTOH I don't believe in python with types yet, so nothing of this really exists for me? I guess?

PR #369, opened 2022-04-29
Re-pin some dependencies

exarkun:
Fixes #286


meejah (approved):
looks fine to me
(note it'll change some formatting because black upgraded, so maybe worth doing that in this PR?)

exarkun:
> looks fine to me (note it'll change some formatting because black upgraded, so maybe worth doing that in this PR?)

Oh, indeed..  I thought I had those changes but I guess I fumbled them somewhere...

Issue #371, opened 2022-05-06
Argument binding is sensitive to placeholder characters that do not really represent placeholders


exarkun:
`_zkapauthorizer.sql.bind_arguments` cannot tell if a question mark represents a placeholder or is appearing in a different kind of token in the SQL statement.  For example, this statement to select the string `"?"`

> SELECT '?'

and this statement to select column `a` from a table named `?`

> SELECT a FROM "?"

will confuse it.

So far this doesn't matter because none of the SQL ZKAPAuthorizer needs to execute uses a question mark for anything other than a placeholder.  Also, `bind_arguments` tries to notice if the input string is something it cannot handle and fail fast rather than producing an incorrect result.

If we wanted event stream construction to be complete and correct it might be better to give up on the idea of manual argument binding and serialize the statement and arguments separately (which should be straightforward given our use of CBOR for the serialization).

The original argument for producing a SQLite3 CLI-compatible snapshot & event stream was compelling but it was based on an assumption that the tooling would make implementation more straightforward than it has proven.  Also, we have no near-term plans to actually run any of this data through the SQLite3 CLI.  Also, if we need to, we can tackle the argument binding problem then and write a converter from another format we choose to a SQLite3 CLI-compatible format.

PR #373, opened 2022-05-09
Change all datetimes to timezone-aware UTC values

exarkun:
This is related to #371 .  The existing `EventStream` code avoids serializing datetime instances using cbor2 by formatting them into the bound SQL statement string.  This lets it avoid the cbor2 behavior of attempting to serialize naive datetimes - which is to refuse to do so by throwing an exception.

After this change, datetimes are all in UTC and timezone aware (including those which may already be in databases out there).  This will make it easier to change the `EventStream` serialization code to serialize datetime instances.

Also, by switching the in-database datetimes to UTC it makes them unambiguous and prevents losing further data about the timestamps they represent.

### DIFF
--- a/src/_zkapauthorizer/tests/test_schema.py
+++ b/src/_zkapauthorizer/tests/test_schema.py
@@ -17,14 +17,45 @@
 Tests for ``_zkapauthorizer.schema``.
 """
 
+from datetime import datetime, timezone
+from typing import Optional
+
+from fixtures import TempDir
+from hypothesis import assume, given
+from hypothesis.strategies import SearchStrategy
+from hypothesis.strategies import datetimes as naive_datetimes
+from hypothesis.strategies import integers, none, one_of
 from testtools import TestCase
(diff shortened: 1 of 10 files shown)


src/_zkapauthorizer/tests/test_schema.py
```diff
@@ -17,18 +17,130 @@
...
+    Build naive datetime instances that represent times that actually happened
+    or will actually happen bounded within a pretty reasonable range around
+    when this code was written.
+    """
+    return naive_datetimes(
+        allow_imaginary=False,
+        # The software has not existing very long.  There is no point going
+        # very far back in time.  Also, according to authorities, timezones
+        # are basically meaningless before 1900 anyway.
+        min_value=datetime(1995, 1, 1, 0, 0),
```
meejah, line 46: good vintage ;)

meejah (approved):
:+1:

PR #377, opened 2022-05-13
Replicate snapshots

### LINKED ISSUE #374: Periodically upload new snapshots
Related to #235 

The service that uploads event-stream data should also upload snapshots sometimes (and prune old event-stream data).

meejah:
Relates to #374

### DIFF
--- a/src/_zkapauthorizer/replicate.py
+++ b/src/_zkapauthorizer/replicate.py
@@ -61,10 +61,10 @@ __all__ = [
 
 import os
 import re
+import sqlite3
 from io import BytesIO
 from sqlite3 import Connection as _SQLite3Connection
 from sqlite3 import Cursor as _SQLite3Cursor
-import sqlite3
 from typing import (
     Any,
     Awaitable,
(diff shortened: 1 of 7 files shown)


meejah:
Is something up with circle-ci? Last few pushes it seems like the main tests (e.g. "linux tests, python 3.9") just never run...? (But, other CircleCI things are running)

meejah:
@exarkun should be good for review

src/_zkapauthorizer/replicate.py
```diff
@@ -623,9 +664,10 @@ def from_statements(
         # although the statement text should still dominate.
         return cls(important, sum(map(len, bound_statements)))
 
-    def __add__(self, other):
+    def __add__(self, other) -> AccumulatedChanges:
+        # if below args are backwards, mypy doesn't seem to complain; why?
```
meejah, line 668: adding the `-> AccumulatedChanges` didn't help, but i left it in...

src/_zkapauthorizer/tahoe.py
```diff
@@ -404,7 +442,16 @@ def client(self, basedir: Optional[FilePath] = None) -> ITahoeClient:
...
+        def encode(s: str) -> str:
+            return b32encode(s.encode("ascii")).decode("ascii")
+
+        cap = "URI:CHK:{}:{}:{}:{}:{}".format(
+            encode("{:016}".format(self._counter)),
+            encode("{:032}".format(self._counter)),
+            self._counter,
+            self._counter,
+            self._counter,
+        )
```
meejah, line 454: I had to add properly-formatted URIs here because of the way the test works for "try to delete a thing from not-a-directory" .. could change the test, I guess, but proper immutable URIs seems like a win.

PR #382, opened 2022-05-18
Take the first snapshot

### LINKED ISSUE #379: Take a snapshot as soon as replication is turned on
There is probably some data in the database that needs to be represented as a snapshot for subsequence `event-stream` data to build upon.

exarkun:
Fixes #379 

Rather than thread a third replica-related argument through the various layers, I wrapped the new "list directory entries" up with the existing uploader and pruner in a single `Replica` type (still preserving the possibility of non-Tahoe-LAFS-based replicas).

I also moved a lot of repetitive test code for creating a `TemporaryVoucherStore` with an empty Tahoe configuration into `TemporaryVoucherStore` itself by giving it a default empty configuration and letting most tests get that value.

I also adjusted the existing `test_enable_replication_on_connection` to make an assertion about behavior instead of internal implementation details.

The actual feature is implemented with a new `startup` job type that is queued first, once, when the replication service starts.

I noticed there is a test missing for the `should_upload_eventstream` logic that was already present (and is moved by this PR).  I'll add that test in a follow-up.

Other follow-up work will be making `should_upload_snapshot` smarter: it should decide yes if a new snapshot would be much smaller than the on-grid footprint of the stored snapshot and all of its event stream objects.  Also, periodic re-checking of this condition (perhaps after each event-stream object upload).

### DIFF
--- a/src/_zkapauthorizer/tests/test_replicate.py
+++ b/src/_zkapauthorizer/tests/test_replicate.py
@@ -10,8 +10,11 @@ from sqlite3 import OperationalError, ProgrammingError, connect
 from typing import BinaryIO, Callable, Optional
 
 from attrs import frozen
+from hypothesis import given
+from hypothesis.strategies import lists, text
 from testtools import TestCase
 from testtools.matchers import (
+    Contains,
     Equals,
     HasLength,
     MatchesListwise,
@@ -30,8 +33,11 @@ from ..model import RandomToken, aware_now
 from ..recover import recover
 from ..replicate import (
     EventStream,
+    Replica,
     get_events,
+    get_tahoe_lafs_direntry_lister,
     get_tahoe_lafs_direntry_pruner,
+    get_tahoe_lafs_direntry_replica,
     replication_service,
     snapshot,
     with_replication,
@@ -317,6 +323,24 @@ class _MatchUpload(Matcher):
         return None
 
 
+async def noop_upload(name, get_bytes) -> None:
+    pass
+
+
+async def noop_prune(predicate) -> None:
+    pass
+
+
+async def noop_list() -> list[str]:
+    return []
+
+
+# A replica that actually does nothing.  Used by tests that don't interact
+# with this part of the replication system but still need a value of the right
+# type.
+noop_replica = Replica(noop_upload, noop_prune, noop_list)
+
+
 class ReplicationServiceTests(TestCase):
     """
     Tests for ``_ReplicationService``.
@@ -336,13 +360,7 @@ class ReplicationServiceTests(TestCase):
         # store.start_lease_maintenance().finish()
         # self.assertThat(get_events(store._connection).changes, HasLength(0))
 
-        async def uploader(name, get_bytes):
-            pass
-
-        async def pruner(predicate):
-            pass
-
-        service = replication_service(store._connection, uploader, pruner)
+        service = replication_service(store._connection, noop_replica)
         service.startService()
         self.addCleanup(service.stopService)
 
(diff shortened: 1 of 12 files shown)


src/_zkapauthorizer/tests/test_replicate.py
```diff
@@ -318,240 +327,284 @@ def match(self, matchee: tuple[str, Callable[[], BinaryIO]]) -> Optional[Mismatc
...
-
-        uploads = []
-        upload_completed = False
-
-        async def uploader(name, get_data):
-            uploads.append((name, get_data))
-            await wait_d
-            nonlocal upload_completed
-            upload_completed = True
+        delay_controller, delay_client = delayedProxy(
```
exarkun, line 457: Using the `delay_controller` to exercise the precise cases we want to exercise in the tests sometimes works out nicely but sometimes ends up pretty gross, with some hard-coded number of `run` calls required to make sure the internal state ends up in just the right place.

I think an improvement would be to write these as Hypothesis stateful tests where "an async thing you started has completed" is one possible transition for the state machine.  This would let Hypothesis explore the stateful space for us in one general test instead of requiring us to hand-craft careful transitions to just the right case we were interested in for one specific test.

meejah: My attempt at writing a hypothesis stateful-thing for magic-folder failed .. so either I don't understand their docs yet or tried something too ambitious :)

src/_zkapauthorizer/tests/test_replicate.py
```diff
@@ -318,240 +327,284 @@ def match(self, matchee: tuple[str, Callable[[], BinaryIO]]) -> Optional[Mismatc
...
 
-    def test_replicate(self) -> None:
-        """
-        Making changes to the voucher store while replication is turned on
-        causes event-stream objects to be uploaded.
-        """
 
-        def get_config(rootpath, portnumfile):
-            return EmptyConfig(FilePath(rootpath))
+def repeatUntil(condition: Callable[[], bool], action: Callable[[], object]) -> None:
```
meejah, line 356: probably `snake_case` to match rest..?

src/_zkapauthorizer/tests/test_replicate.py
```diff
@@ -318,240 +327,284 @@ def match(self, matchee: tuple[str, Callable[[], BinaryIO]]) -> Optional[Mismatc
...
-        def add_tokens():
-            tokens = [RandomToken(b64encode(urandom(96))) for _ in range(10)]
-            voucher = urlsafe_b64encode(urandom(32))
-            tvs.store.add(voucher, len(tokens), 1, lambda: tokens)
+        # We'll spy on "events" which are only captured when the connection is
+        # in replication mode.  To start, make sure that database changes are
+        # not already being captured.  They should not be since nothing has
+        # placed the connection into replication mode yet.
+        # store.start_lease_maintenance().finish()
+        # self.assertThat(get_events(store._connection).changes, HasLength(0))
```
meejah, line 405: not clear if you meant to include these assertions (but I guess: delete them if not)

meejah: (looks like they happen below, so "delete"?)

src/_zkapauthorizer/tests/test_replicate.py
meejah: I guess another disadvantage of the "delay_controller" is that if the test would "normally" fail, it'll just hang now? (because e.g. 3 files never appear)

meejah (approved):
some stuff inline, looks good otherwise

src/_zkapauthorizer/tests/test_replicate.py
exarkun: > I guess another disadvantage of the "delay_controller" is that if the test would "normally" fail, it'll just hang now? (because e.g. 3 files never appear)

Potentially, yea. :(  `delay_controller.run` mitigates this somewhat by raising an exception if there is nothing delayed - so if the failure is that the implementation stops trying to do anything before getting to the desired state, it'll fail quickly.  But if the failure is that the implementation keeps trying to do work but never gets to the desired state, yea, repeat forever.

This could be mitigated by adding a constraint on the maximum number of repeats.  I might prefer instead to spend the effort trying to write the tests so they don't have these sensitivities on implementation details though.

exarkun: > My attempt at writing a hypothesis stateful-thing for magic-folder failed .. so either I don't understand their docs yet or tried something too ambitious :)

I _hope_ this will be possible for the replication service but I'm not completely sure I already know how to do it.  So we'll see...

src/_zkapauthorizer/tests/test_replicate.py
exarkun: Oops.  Good thing the test suite still passes with these in.  The ones below invert the condition (`Not(HasLength(0))`).  These are just meant to prove some assumptions the test is making holds (that there are no surprise changes in the database before the test even tries to do anything).

PR #386, opened 2022-05-20
Test event-stream-upload-at-start behavior of replication service

### LINKED ISSUE #384: Replication service's behavior of uploading local event stream information at startup is not tested
The actual process of uploading an event stream is tested but the behavior of triggering this at startup is not.

exarkun:
Fixes #384


src/_zkapauthorizer/tests/test_replicate.py
```diff
@@ -385,6 +385,16 @@ def download_event_stream(cap):
     )
 
 
+@log_call(action_type="zkapauthorizer:tests:add-tokens")
+def add_tokens(store: VoucherStore) -> None:
```
exarkun, line 389: Moved out of an existing test for re-use

PR #387, opened 2022-05-20
Switch from netstring to CBOR

### LINKED ISSUE #378: encode snapshots using CBOR instead of netstrings
We used CBOR for the event stream.  Let's be consistent between these two things.

CBOR is a better choice than netstrings because it supports types other than strings.

exarkun:
Fixes #378


exarkun:
The Windows error is #389

PR #391, opened 2022-05-23
update Tahoe-LAFS master dependency

exarkun:
(should be) periodic update to catch incompatibilities early


hacklschorsch (approved):
LGTM!

exarkun:
Hmm, did not look good to CI though.  Looks like a minor incompatibility with Tahoe-LAFS master@HEAD.  I think that change was just introduced a couple days ago. We can probably accommodate both old and new versions of this codepath.

PR #392, opened 2022-05-23
drop 21.05 and bump 21.11

exarkun:
(should be) periodic nixpkgs update to get security fixes and notice breakage earlier

we haven't been using the 21.05 pin for a while and we shouldn't start in the future, so also delete it


meejah (approved):
i don't know the implications of any of this but I guess someone needs to push the button :)

exarkun:
Since CI is happy, the implications must all be good, right?  :slightly_smiling_face: 

Basically we have learned that ZKAPAuthorizer's automated test suite still works with the newer versions of packages now present in the NixOS 21.11 release branch.  This is kind of what we would learn every day (whether we like it or not) from CI with unpinned dependencies (like if we just ran `pip install ...` on Ubuntu for each job).

PR #407, opened 2022-05-27
Make and upload a new snapshot sometimes

### LINKED ISSUE #380: Take a new snapshot when doing so is advantageous with respect to storage efficiency
The replica state consists of a fixed-size snapshot and an ever-growing collection of event streams.  Thus, the replica state grows in size indefinitely.

This is mitigated by periodically taking a new snapshot, basing all new event streams on it, and pruning the old snapshot and all of its event stream.

This bounds the size of the replica to the useful data in the database since the new snapshot will collapse the event-stream replication of the add/delete pairs for any data that no longer exists.

At some point, the size of the existing snapshot and event stream grows large enough that it is worth taking a new snapshot and uploading it.  The system should do this automatically at that point.

exarkun:
Fixes #380

### DIFF
--- a/src/_zkapauthorizer/_plugin.py
+++ b/src/_zkapauthorizer/_plugin.py
@@ -31,7 +31,7 @@ from allmydata.interfaces import (
     IFoolscapStoragePlugin,
 )
 from allmydata.node import MissingConfigEntry
-from attrs import Factory, define, field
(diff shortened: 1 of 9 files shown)


exarkun:
Some test coverage is missing, working on that now.

PR #410, opened 2022-06-01
WebSocket-based restore

meejah:
@exarkun 
The client side of this might look like:

```
import json
from autobahn.twisted.websocket import (
    create_client_agent,
)
from twisted.internet import reactor
from twisted.internet.defer import inlineCallbacks, returnValue, Deferred


@inlineCallbacks
def main(reactor):
    agent = create_client_agent(reactor)
    uri = "ws://127.0.0.1:8888/storage-plugins/privatestorageio-zkapauthz-v2/recover"
    dircap = "URI:DIR2-RO:oibl7ox3nc4cwjqx4f3nrwquxi:qjbyvkxqey7lyaktni6gbg5zyzeiphfnk6jdvjrp6y4wf2ktufeq"
    print("URI", uri)
    with open("/home/meejah/.config/gridsync/PrivateStorage Staging/private/api_auth_token") as f:
        token = f.read()
    proto = yield agent.open(
        uri,
        {"headers": {"Authorization": f"tahoe-lafs {token}"}},
    )

    def status_update(data, is_binary=False):
        print(f"status update: {data}")
        # XXX e.g. hook into UI to display state
    proto.on("message", status_update)

    print("wait for open")
    yield proto.is_open
    print("send recovery request")
    yield proto.sendMessage(json.dumps({"recovery-capability": dircap}).encode("utf8"))
    try:
        print("wait for close")
        yield proto.is_closed
    except Exception as e:
        raise TahoeWebError(f"Error during recovery: {e}")



from twisted.internet.task import react
react(main)
```


meejah:
Tests added

meejah:
besides "there's something up with dependencies" which I have yet to figure out, i think this is ready for review @exarkun 

(I think the trick is to find an Autobahn with the right testing fixes that is old enough to keep the rest happy?)

default.nix
```diff
@@ -33,6 +33,8 @@ in
       # This is kind of round-about but it seems to be the best way to
       # convince mach-nix to use a specific package for a specific dependency.
       tahoe-lafs = "nixpkgs";
+
+      cryptography = "nixpkgs";
```
exarkun, line 37: The packaging fix was mostly this.  mach-nix can't find a new enough cryptography to satisfy the newer autobahn using only the "sdist" and "wheel" providers specified above.  "nixpkgs" lets it get one that is just barely new enough (but still quite old, it would be nice to fix the issue that prevents it from getting a more recent version...)

src/_zkapauthorizer/recover.py
```diff
@@ -104,6 +104,7 @@ class StatefulRecoverer:
     """
 
     _state: RecoveryState = RecoveryState(stage=RecoveryStages.inactive)
+    _listeners: Iterable[Callable[[RecoveryState], object]] = field(default=())
```
exarkun, line 107: I made listeners an initialization-time thing and removed the API for adding more of them later.  This was enough for all consumers of the interface and removes the possibility for weird listeners-changed-during-broadcast behavior.

src/_zkapauthorizer/resource.py
```diff
@@ -214,110 +213,174 @@ def _setup_replication(self, request):
...
+        """
+        try:
+            body = loads(payload)
+            if set(body.keys()) != {"recovery-capability"}:
+                raise ValueError("Unknown keys present in request")
+            recovery_capability = from_string(body["recovery-capability"])
+            if not isinstance(recovery_capability, ReadonlyDirectoryURI):
+                raise ValueError("Not a readonly-dircap")
+        except Exception:
+            self.sendClose(
```
exarkun, line 255: It would be nice to have some error logging here in case the exception is something surprising.

src/_zkapauthorizer/resource.py
```diff
@@ -214,110 +213,174 @@ def _setup_replication(self, request):
...
+    no matter how many clients there are and because something needs
+    to link to other resources that are also constructed once.
+    """
 
     store: VoucherStore = field()
     get_downloader: Callable[[str], Downloader] = field()
-    recoverer: StatefulRecoverer = field(default=Factory(StatefulRecoverer))
+    recoverer: StatefulRecoverer = field()
+    recovering_d: Deferred = field(default=None)
+    recovering_cap: CapStr = field(default=None)
```
exarkun, line 278: in `initiate_recovery` `recovering_cap` is set to a `ReadonlyDirectoryURI`... FWIW it looks like `ReadonlyDirectoryURI` is the type actually used at runtime.

Also, these two recovering fields should actually be `Optional[...]`?  I dunno wtf is wrong with mypy.

Does subclassing something from autobahn just break all mypy inference for this whole class?

src/_zkapauthorizer/resource.py
```diff
@@ -214,110 +213,174 @@ def _setup_replication(self, request):
...
+            body = loads(payload)
+            if set(body.keys()) != {"recovery-capability"}:
+                raise ValueError("Unknown keys present in request")
+            recovery_capability = from_string(body["recovery-capability"])
+            if not isinstance(recovery_capability, ReadonlyDirectoryURI):
+                raise ValueError("Not a readonly-dircap")
+        except Exception:
+            self.sendClose(
+                code=4000,
+                reason="Failed to parse recovery request",
```
exarkun, line 257: The previous implementation revealed more information about what the client has done wrong.  It would be nice to preserve that.

src/_zkapauthorizer/resource.py
```diff
@@ -214,110 +213,174 @@ def _setup_replication(self, request):
...
-            request.write(b"there is existing local state")
-            request.finish()
-        except:
-            # And if something else is broken, then who knows...  At least try
-            # to generate an error response.
-            self._log.failure("recovery setup failed")
-            request.setResponseCode(INTERNAL_SERVER_ERROR)
-            request.write(b"recovery setup failed")
-            request.finish()
+            raise ValueError("there is existing local state")
```
exarkun, line 382: Perhaps worth just letting `NotEmpty` through as well since now there is an exception handler wrapped around this whole coroutine.

src/_zkapauthorizer/resource.py
```diff
@@ -648,3 +711,33 @@ def bad_request(reason="Bad Request"):
         b"Bad Request",
         reason.encode("utf-8"),
     )
+
+
+async def recover(
```
exarkun, line 716: I added this not-extremely-useful client helper API because 3 or 4 tests were essentially duplicating it.

src/_zkapauthorizer/tests/test_client_resource.py
```diff
@@ -719,203 +751,208 @@ def create(store, state):
...
+        downloads = []
+        downloading_d: Deferred[None] = Deferred()
+
+        def get_success_downloader(cap):
+            async def do_download(set_state):
+                await downloading_d
+                downloads.append(set_state)
+                return (
+                    # this data is CBOR for {"version": 1, "statements": []}
+                    lambda: BytesIO(b"\xa2gversion\x01jstatements\x80"),
```
exarkun, line 898: This could be `statements_to_snapshot([])`.

src/_zkapauthorizer/tests/test_plugin.py
```diff
@@ -840,42 +846,45 @@ def test_downloader(self, get_config):
...
-                )
-            ),
+
+        # normally, we'd use Treq's machinery to do in-memory
+        # requests, but this is a WebSocket, so we want to use
+        # Autobahn's testing machinery. Maybe there's a clever way to
+        # hook those together, but for now we reach in "directly" to
+        # grab the WebSocketResource and set up Autobahn's test agent
+        # that way (requiring the factory from the real resource)...
+        wsr = root._portal.realm._root.children[b"recover"]
```
exarkun, line 856: I would like to find a way to avoid doing this.

exarkun (commented):
Thanks.  I pushed some changes and also left some inline comments.

src/_zkapauthorizer/tests/test_client_resource.py
meejah: Did think of that ... would that be better, or better to have "concrete data that doesn't change"...?

PR #425, opened 2022-08-19
Further changes towards a clean strict mypy report

exarkun:


hacklschorsch (approved):
I skipped over this; what I saw looked good to me.
It's too much code for me to "read" and grokk in a useful way really.

Trying to execute mypy via flakes I currently get:

 ```
$ nix --experimental-features 'nix-command flakes' run .#mypy -- src
Success: no issues found in 63 source files
```

Which seems to be doing what it should.

exarkun:
> I skipped over this; what I saw looked good to me. It's too much code for me to "read" and grokk in a useful way really.

"Skimmed", maybe? :slightly_smiling_face: 

> 
> Trying to execute mypy via flakes I currently get:
> 
> ```
> $ nix --experimental-features 'nix-command flakes' run .#mypy -- src
> Success: no issues found in 63 source files
> ```
> 
> Which seems to be doing what it should.

Thanks for having a look @hacklschorsch .  I agree this is too large to be reviewable in any meaningful way.  Originally I thought that I might try to split it up after getting to the point where mypy reported no errors.  I _could_ do that but I think it would take a while.  The changes would need to be landed in some kind of dependency order (roughly a topological ordering of dependencies between types as well as between the expressions that use them).  Even computing this order is probably a large project.  

So instead I re-reviewed the parts of the branch that could actually change runtime behavior and I'm satisfied that _probably_ those changes _mostly_ improve failure modes.  With this, together with your review and green CI (pending ...), I feel okay merging this.

PR #430, opened 2022-09-09
Switch ITahoeClient to use tahoe-capabilities instead of str and home-grown functionality

### LINKED ISSUE #427: Use APIs from tahoe-capabilities instead of `str` and APIs from Tahoe-LAFS for work on capabilities
This is an umbrella ticket.  There are likely many self-consistent, increment steps that can be taken towards the overall goal.  These can be described in separate tickets and resolved individually.

ZKAPAuthorizer works on capabilities in many places.  It frequently represents these capabilities using `str` or `bytes` values.  These are problematic because they are opaque:

* is any given `str` or `bytes` value a capability?  Who knows.
* What kind of capability does a given `str` or `bytes` value represent, if any?  You won't know until you parse it.

Additionally, where more expressive types are used they are frequently from Tahoe-LAFS which doesn't promise a public Python API.

Instead of these things, use the recently created and released `tahoe-capabilities` library.  Where that library is missing functionality we need to get away from string representations and private Tahoe-LAFS APIs, add that functionality to `tahoe-capabilities`.

exarkun:
Fixes #427

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -85,6 +85,9 @@ show_column_numbers = True
 pretty = True
 show_error_codes = True
 
+[mypy-_zkapauthorizer.tests.test_tahoe]
+disallow_subclassing_any = False
+
 [mypy-_zkapauthorizer.tests.*]
 disallow_untyped_calls = False
 disallow_subclassing_any = False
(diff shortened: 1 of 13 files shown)


hacklschorsch:
Belated review: I think his is a lovely patch.

PR #438, opened 2022-09-20
Use the simpler `cachix watch-exec` to populate our cachix cache

exarkun:
While setting up cachix for PrivateStorageio/PaymentServer I noticed some things could be simpler.  Here they are.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -316,16 +316,20 @@ jobs:
     environment:
       <<: *NIX_ENVIRON
 
-      # CACHIX_AUTH_TOKEN is manually set in the CircleCI web UI and allows us to push to CACHIX_NAME.
+      # CACHIX_AUTH_TOKEN is manually set in the CircleCI web UI and allows us
+      # to push to CACHIX_NAME.
       CACHIX_NAME: "privatestorage-opensource"
 
     steps:
       - run: &SETUP_CACHIX
           name: "Set up Cachix"
           command: |
+            # Install cachix, the Nix-friendly caching tool.
             nix-env -f $NIXPKGS -iA cachix bash
+            # Activate it for "binary substitution".  This sets up
+            # configuration that lets Nix download something from the cache
+            # instead of building it locally, if possible.
             cachix use "${CACHIX_NAME}"
-            nix path-info --all > /tmp/store-path-pre-build
 
       - "checkout"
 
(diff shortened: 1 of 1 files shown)


hacklschorsch:
I don't know how to implement cachix  - so probably these would be answered by going through the docs - I'll still ask since I feel lazy right now:

1. The `build-wheel` job seems to **not** use cachix. Is that intentional?
2. The `linux-tests` job uses cachix, and seems to also store build artifacts.  But do we really want to cache test results?  Can we (easily) / should we maybe only cache build results but not test results?
3. The `restore_cache` and `save_cache` steps in the `macos-tests` job - are they valid for cachix?

hacklschorsch (approved):
Seems like a clear improvement!  Please also see my [comment](https://github.com/PrivateStorageio/ZKAPAuthorizer/pull/438#issuecomment-1253490276).

exarkun:
> I don't know how to implement cachix - so probably these would be answered by going through the docs - I'll still ask since I feel lazy right now:
> 
>     1. The `build-wheel` job seems to **not** use cachix. Is that intentional?

I think that was accidental.  However, I don't think the wheel job will benefit much from cachix.  Unlike the test jobs it doesn't have to build any of the Python dependencies.  Python itself and other similar dependencies will be retrieved from the NixOS cache.  The wheel _itself_ could be cached on cachix but it will usually different so the cache won't be usable and "building" it only takes a few seconds anyway (it's just a zip archive of the source).

I'm okay leaving this uncached for now and if we notice the wheel job is taking too long we can revisit this idea.

> 
>     2. The `linux-tests` job uses cachix, and seems to also store build artifacts.  But do we really want to cache test results?  Can we (easily) / should we maybe only cache build results but not test results?

cachix itself doesn't give us any way to do this filtering so I don't think it will be easy to change this behavior.  I think caching and not caching each have arguments in their favor here so I'm happy to stick with the easy option until we develop stronger feelings.

> 
>     3. The `restore_cache` and `save_cache` steps in the `macos-tests` job - are they valid for cachix?

Unfortunately we can only use cachix for the Nix builds.  So the `macos-tests` job still uses CircleCI's built-in caching features.  I think that's probably okay, at least for a while.  The `macos-tests` job's cache doesn't seem to be growing unreasonably - probably because it only ends up caching the Python wheels we explicitly put into the "wheelhouse" directory.  This will grow as new versions of our dependencies are released, but probably very slowly.  Right now the cache appears to be 30MB and takes 3s to restore.

PR #440, opened 2022-09-20
Avoid double-redeeming vouchers

### LINKED ISSUE #437: UNIQUE constraint failed: unblinded-tokens.token
Apparently redemption can result in this exception:

```
[_zkapauthorizer.controller.PaymentController#critical] Redeeming random tokens for a voucher (b'...') encountered error.
     File "twisted\internet\defer.py", line 1791, in gotResult

     File "twisted\internet\defer.py", line 1781, in _inlineCallbacks

     File "twisted\internet\defer.py", line 695, in callback

     File "twisted\internet\defer.py", line 797, in _startRunCallbacks

   --- <exception caught here> ---
     File "twisted\internet\defer.py", line 891, in _runCallbacks

     File "_zkapauthorizer\controller.py", line 933, in _redeem_success

     File "_zkapauthorizer\model.py", line 223, in with_cursor

     File "_zkapauthorizer\model.py", line 565, in insert_unblinded_tokens_for_voucher

     File "_zkapauthorizer\replicate.py", line 538, in executemany

   sqlite3.IntegrityError: UNIQUE constraint failed: unblinded-tokens.token

```

It may be transient.  It's not entirely clear to me how this case can arise, though.

exarkun:
Fixes #437 

Before we would try to redeem a voucher concurrently if someone asked us to but only one of these attempts could ever succeed.  Now we'll notice if we're already redeeming in the cases that would have led to concurrency and just log the fact instead.

### DIFF
--- a/src/_zkapauthorizer/controller.py
+++ b/src/_zkapauthorizer/controller.py
@@ -929,11 +929,11 @@ class PaymentController(object):
             if voucher_obj.state.should_start_redemption():
                 counter_start = voucher_obj.state.start_at_counter()
             else:
-                raise ValueError(
-                    "Cannot redeem voucher in state {}.".format(
-                        voucher_obj.state,
-                    ),
+                self._log.info(
+                    "Cannot redeem voucher in state {state}.",
+                    state=voucher_obj.state,
                 )
+                return None
 
         self._log.info(
             "Starting redemption of {voucher}[{start}..{end}] for {num_tokens} tokens.",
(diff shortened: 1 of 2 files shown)


src/_zkapauthorizer/controller.py
```diff
@@ -929,11 +929,11 @@ async def redeem(self, voucher: bytes, num_tokens: Optional[int] = None) -> None
             if voucher_obj.state.should_start_redemption():
                 counter_start = voucher_obj.state.start_at_counter()
             else:
-                raise ValueError(
-                    "Cannot redeem voucher in state {}.".format(
-                        voucher_obj.state,
-                    ),
+                self._log.info(
```
meejah, line 932: Could maybe argue for "warning" ... but also: levels, meh.

meejah (approved):
LGTM
(I'll trade you for a review on the tahoe thing :)

PR #441, opened 2022-10-03
use the 1.18 series

### LINKED ISSUE #442: tahoe-lafs 1.18.0
Since tahoe-lafs 1.18.0 is released we should switch to it (assuming there's no issues w/ zkapauthorizer in combination with that release).

meejah:
fixes #442

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -472,7 +472,7 @@ workflows:
     - "documentation"
     - "typecheck":
         py-version: "39"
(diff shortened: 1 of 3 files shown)


exarkun:
I suppose we could make cachix optional, and skip it if the API key isn't found in the environment.

meejah:
Cool thanks, don't really know what I'm doing anyway ;) and not sure if my Nix setup is .. useful? (the `niv` install instructions didn't work for me yet)

PR #443, opened 2022-10-04
Release 22.8.20.1

exarkun:
This widens the allowed range of Tahoe-LAFS versions and bumps the version number.  If it passes CI then we can tag it as a release.  This will make integration with newer Tahoe-LAFS possible without pulling in any of the other recent development work on ZKAPAuthorizer (which is good because it has some known bugs).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -498,6 +498,7 @@ workflows:
               - "310"
             tahoe-lafs:
               - "1_17_1"
+              - "1_18_0"
               # This is usually not master@HEAD because it is still pinned to
               # a certain revision.  The intent is to update it frequently and
               # discover fixable incompatibilities in small groups and
--- a/flake.lock
+++ b/flake.lock
@@ -85,11 +85,11 @@
     "tahoe-lafs-dev": {
       "flake": false,
       "locked": {
-        "lastModified": 1660571400,
-        "narHash": "sha256-atIr+RwV4zSejX0sPqgqKTR2lNvjI2+H5do6lhrslSk=",
+        "lastModified": 1664893933,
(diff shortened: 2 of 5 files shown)


meejah:
so is codecov revealing heisen-coverage, or just confused yet again?

exarkun:
> so is codecov revealing heisen-coverage, or just confused yet again?

It could be heisen-coverage.  The way Hypothesis is set up right now does not guarantee the exact same codepaths will be covered on each CI run.  The plan for this is to make the PR-triggered run use a CI profile that's deterministic and run a, say, nightly CI job that's randomized.  This way PRs don't get the random noise but Hypothesis still has a chance to discover new problems for us.

exarkun:
I had to change the version number from `v2022.8.20.1` to `v2022.8.21` because the release automation doesn't trigger for 4 component version numbers. :(  (And I didn't want to mess with the release automation in this PR)

meejah:
I don't see a problem with the 3-number one (but I guess you're trying to encode that it's a "spot release" on .8.20 ...?)

exarkun:
> I don't see a problem with the 3-number one (but I guess you're trying to encode that it's a "spot release" on .8.20 ...?)

Yea, that.  The biggest annoyance is just another 40 minute CI run, I think.

meejah:
I guess "calver" is somewhat vague; I've used `YY.MM.##` (where `##` just increments) because the most common reason I've had for "another minor release" is that I screwed something up so it's often on the same day ;)

From calver.org's "case studies", looks pretty split on whether the second number is "month" or "a number that goes up".

I suppose to support a use-case like this, it would make sense for the second number to just be "something that increases" (so then "normally" you'd increase that number, but for patch/spot/whatever releases, the right-most number would increase). Or, use 4 numbers .. but that seems far less common IME.

PR #446, opened 2023-02-16
Un-mach

hacklschorsch:
`mach-nix` unfortunately is broken with this package in NixOS 22.11.  This is an effort to rip out mach-nix and use Nix's own `buildPythonPackage` instead.


nix/lib.nix
```diff
@@ -27,18 +27,24 @@ rec {
...
+	pname = "ZKAPAuthorizer";
+        version = "9001";
+        propagatedBuildInputs = [
+          prometheus-client
+          colorama
+          tahoe-lafs-package
+          (callPackage ./compose.nix {})
+          (callPackage ./tahoe-capabilities.nix {})
+          sqlparse
+          autobahn
```
hacklschorsch, line 46: This is where we stopped last - `nix build` isn't happy about the version of autobahn we have in this nixpkgs.

hacklschorsch:
Today we stopped at trying to build pycddl as an input to tahoe-lafs as an input to ZKAPAuthorizer

hacklschorsch:
This is actually fit for a review now - I removed the "Draft/WIP" words in the title - but some tests are failing and need to be fixed.

exarkun:
I filed https://github.com/HypothesisWorks/hypothesis/issues/3630 upstream for the test suite failures.  I don't really understand what's going on with Hypothesis here.

PR #449, opened 2023-03-30
Add flake-compat

hacklschorsch:
Copy-pasting from https://github.com/edolstra/flake-compat so we can use this from PrivateStorageio even when that's not a flake yet.


exarkun:
The Windows failure is fixed by #450 .  The macOS failure looks harder and I guess we'll just live with it for now.

PR #450, opened 2023-03-30
Fix Python 3.10 Windows CI

exarkun:
The Windows image for the Windows CI environment was changed at the beginning of March 2023.  The new image does not have Python 3.10, only Python 3.9 and Python 3.11.

The Windows Python 3.10 CI job therefore failed as `py` could not find a Python 3.10 to use.

This pins the Windows image to the September 2022 Windows image which does have Python 3.10.

Later we can also pin a newer image if we want to test against Python 3.11.

This does not try to fix the Python 3.9 macOS CI job which seems to fail for unrelated reasons now.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -86,6 +86,7 @@ jobs:
       - run:
           name: "Setup Environment"
           command: |
+            py --list
             py -<< parameters.py-version >> -V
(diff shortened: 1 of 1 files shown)


hacklschorsch (commented):
👍

PR #452, opened 2023-04-27
fix price computation

### LINKED ISSUE #455: PriceCalculator is incorrect
For certain values of inputs, it seems that `PriceCalculator` produces incorrect output.

For example, it doesn't account for the fact that "5 total shares" means that you must upload to 5 different servers, and so therefor the minimum number of ZKAPs spent must be 5 (or more, if the share size exceeds the pass-value).

meejah:
It seems that the number of ZKAPs are under-estimated in the PriceComputation class.

Fixes #455


meejah:
Not sure what's up with the CI, all the nix stuff seems to fail? (The tests pass locally)

exarkun:
> Not sure what's up with the CI, all the nix stuff seems to fail? (The tests pass locally)

Looks a bit like a Twisted bug :/  The nix jobs are all using Twisted 22.8.  I know there were a couple trial fixes in Twisted 22.10.  I wonder if that's what makes the difference.

meejah:
Oh, at least one looks like hypothesis found a thing ...

hacklschorsch (approved):
Big step in the right direction!

src/_zkapauthorizer/tests/test_pricecalculator.py
```diff
@@ -222,3 +223,65 @@ def test_one_pass(self, pass_value: int) -> None:
...
+            shares_needed=needed,
+            shares_total=needed + extra_shares,
+        )
+        price = calculator.calculate([1000])
+        self.assertThat(price, greater_or_equal(needed))
+
+    @given(
+        integers(min_value=1, max_value=100).flatmap(
+            lambda n_shares: lists(sizes(), min_size=n_shares, max_size=n_shares)
+        ),
```
exarkun, line 253:
```suggestion
        lists(sizes(), min_size=1, max_size=100),
```

If `lists` only supported a `size=...` argument then the flatmap construction here would be useful but since it supports the same kind of size bounds as `integers` you don't really need it.

src/_zkapauthorizer/tests/test_pricecalculator.py
```diff
@@ -222,3 +223,65 @@ def test_one_pass(self, pass_value: int) -> None:
             price,
             Equals(1),
         )
+
+    @given(
+        integers(min_value=2, max_value=255),
+        integers(min_value=0, max_value=254),
+    )
+    def test_minimum_spending(self, needed: int, extra_shares: int) -> None:
```
exarkun, line 231: There's also `_zkapauthorizer.tests.strategies.encoding_parameters`.

exarkun (approved):
Thanks.  A couple very minor comments about the tests inline, address and merge when you're satisfied.

PR #466, opened 2024-10-30
draft: attrs/provides experimental (NOT FOR MERGE)

hacklschorsch:
Pinning attrs to an old version

... is not a real solution to our problems,
but it does fix a lot of tests for me,
so I'll push this once to see if it makes all
the Windows tests go green or if there's others
missing.

As I understand it, we can't just keep this 'fix'
since tahoe-lafs will have a newer version of
attrs and we can't have multiple versions of a library in Python.


hacklschorsch:
This served its purpose.

Issue #462, opened 2024-10-30
Some Circle-CI checks are failing and blocking a new release


btlogy:
Possibly related to #454

@crwood has been trying to debug this in #461

![image](https://github.com/user-attachments/assets/3203921d-fd32-45be-8777-afbcf3c2b1ac)

Thus, 5 jobs are still failing:
- [ ] windows-tests-3.9
- [ ] windows-tests-3.10
- [ ] macos-test-3.9-13.4.1
- [ ] macos-test-3.10-13.4.1
- [ ] linux-test-39-dev

btlogy:
@laurih01: both @hacklschorsch and @btlogy would like to have Write/Commit access to this repository, so we could "Rerun jobs with SSH" in Circle-CI.

btlogy:
I've managed to reproduce the same test error locally using `nix build --verbose .#tests-python39-tahoe_1_18_0-ci-cov`:

```
_zkapauthorizer.tests.test_client_resource.RecoverTests.test_conflict id=0x7ffff5091a90> took longer than 60.0 seconds
```

It seems to me that the `ci` profile which requires `max_examples=200` is indeed hitting this timeout of 60s, no matter with which version of `tahoe` we pick.
This [build](https://app.circleci.com/pipelines/github/PrivateStorageio/ZKAPAuthorizer/1887/workflows/1a098f3f-1bea-4327-b437-1a11ceb3cfdf) triggered by @hacklschorsch are confirming that the failure of those `linux-tests` are transient.

laurih01:
> @laurih01: both @hacklschorsch and @btlogy would like to have Write/Commit access to this repository, so we could "Rerun jobs with SSH" in Circle-CI.

Done, you both have maintain access now.

btlogy:
Now I've maintainer access, I've been playing around with CircleCI and I'm not happy with my understanding so far:

No matter if it is a public repo (unless there is a way to pull via https) CircleCI needs a private ssh key to checkout the code from GitHub.

There is apparently 3 ways to provide an [ssh](https://app.circleci.com/settings/project/github/PrivateStorageio/ZKAPAuthorizer/ssh) key to CircleCI (at the project level):

1. CircleCI can conveniently create a private key and use GitHub OAuth to authorize the public part as a deploy key for the related project only. But I suspect this to work only if the acting user has admin permissions on the repo (nothing happens for me).
[...]

PR #467, opened 2024-10-31
Use https instead of ssh to checkout the code from CircleCI

btlogy:
Part of #462

This PR would avoid us having to manage some SSH key(s) only to read a public repo (until we have to push?).
It introduces a re-usable [custom checkout step](https://support.circleci.com/hc/en-us/articles/115015659247-How-do-I-modify-the-checkout-step) using https URL instead of the ssh one provided by GitHub OAuth.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,13 +19,19 @@ orbs:
   win: "circleci/windows@4.1.1"
 
 aliases:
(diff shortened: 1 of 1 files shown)


crwood (approved):

hacklschorsch:
We did try this, yes? https://batsov.com/articles/2022/09/20/resetting-circleci-checkout-ssh-keys/

PR #475, opened 2024-11-01
Force CI to reset the main branch on checkout

btlogy:
Part of #462

This should fix the CI when building the main branch (which is checkout twice).

I'd like to improve this whole custom checkout, but I'll do this in an other project before porting changes here.
This is just a minimal change to get the CI to build.


crwood (approved):
Thanks!

PR #476, opened 2024-11-04
Fix attrs >= 24 missing `provides()`

hacklschorsch:
attrs [deprecated part of their API (`provides()`)](https://github.com/python-attrs/attrs/pull/1265) and we have to deal with it.


hacklschorsch:
This was more discussing with CI than I was hoping for, but now is in reviewable state I'd say @crwood @btlogy

crwood (approved):
Looks good. Thank you!

PR #477, opened 2024-11-04
Pin attrs until we have a compatible tahoe-lafs

hacklschorsch:
This pin can and should be removed when we start
to depend on tahoe-lafs 1.20 or up.


crwood (approved):
Thanks!

PR #479, opened 2024-11-04
Update nixpkgs to current stable channel

hacklschorsch:
To work with latest tahoe-lafs we need newer dependency versions all around.

This PR updates nixpkgs to the current stable channel.


hacklschorsch:
I am feeling [optimistic](http://hintjens.com/blog:106).

PR #480, opened 2024-11-05
Update challenge-bypass-ristretto

hacklschorsch:
Command used: `nix flake lock --update-input challenge-bypass-ristretto`


hacklschorsch:
I am feeling [optimistic](http://hintjens.com/blog:106) and this will only look a bit better together with #479

PR #483, opened 2024-11-07
Stop SyntaxWarnings because of some escaped blank

hacklschorsch:
I see these often at the very beginning when running tests - I removed that backslash, but also, not so sure what I'm doing here.  @crwood is removing these okay?  Why are they there in the first place?
```
    _zkapauthorizer/replicate.py:393: SyntaxWarning: invalid escape sequence '\ '
    _zkapauthorizer/replicate.py:490: SyntaxWarning: invalid escape sequence '\ '
```

### DIFF
--- a/src/_zkapauthorizer/replicate.py
+++ b/src/_zkapauthorizer/replicate.py
@@ -390,7 +390,7 @@ MutationObserver = Callable[[_SQLite3Cursor, Iterable[Mutation]], Callable[[], N
 
 @define
 class _ReplicationCapableConnection:
-    """
+    r"""
     Wrap a ``sqlite3.Connection`` to provide additional snapshot- and
     streaming replication-related features.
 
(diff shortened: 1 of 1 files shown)


crwood (approved):
Ah, yes, this was a change introduced in Python 3.12 (which raises a `SyntaxWarning` for invalid escape sequences instead of a `DeprecationWarning`). See [here](https://docs.python.org/3/whatsnew/3.12.html#other-language-changes) or [here](https://github.com/python/cpython/issues/98401)

The usual/recommended way to fix this is by making the docstring into a "raw" string (i.e., adding an `r` prefix to the to the first line of the docstring by changing `"""...` to `r"""...`); as per [PEP 257](https://peps.python.org/pep-0257/#what-is-a-docstring):

> Use `r"""raw triple double quotes"""` if you use any backslashes in your docstrings. 

Removing it also works, but depending on what's rendering those docstrings (if they're being rendered by anything?), you might see an extra space where there was no extra space before.

hacklschorsch:
Thank you for your review @crwood !

